PR #76, opened 2021-07-06
Feature/document service

medgar-nitric:


adapters/grpc/document_grpc.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, err
+	}
+}
+
+func (s *DocumentServer) Get(ctx context.Context, req *pb.DocGetRequest) (*pb.DocGetResponse, error) {
+	if ok, err := s.checkPluginRegistered(); ok {
+		key := toSdkKey(req.Key)
+		subKey := toSdkKey(req.SubKey)
+
+		if val, err := s.documentPlugin.Get(*key, subKey); err == nil {
```
tjholm, line 66: See comment on reference v. value, we can avoid having to de-reference the primary key in this case if we just take both by reference.

adapters/grpc/document_grpc.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func NewDocumentServer(docPlugin sdk.DocumentService) pb.DocumentServer {
+	return &DocumentServer{
+		documentPlugin: docPlugin,
+	}
+}
+
+func toSdkKey(key *pb.Key) *sdk.Key {
+	if key != nil {
```
tjholm, line 159: Can add Key validation at the adapter layer here (or in a separate method) to universally validate non-nil keys.

.gitignore
```diff
@@ -22,4 +22,7 @@ bin/
 *.coverprofile
 
 # Ignore proto generated interfaces
-interfaces/
\ No newline at end of file
+interfaces/
+
+# GCP integration testing key
+plugins/document/firestore/gcp-firestore-testing.json
```
tjholm, line 28: Is this still needed?

makefile
```diff
@@ -22,6 +22,13 @@ install-tools: install
 	@echo Installing tools from tools.go
 	@cat ./tools/tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go get %
 
+# Install integration testing tools
+install-test-tools:
+	@wget https://s3.us-west-2.amazonaws.com/dynamodb-local/dynamodb_local_latest.tar.gz
+	@sudo mkdir -p /usr/local/dynamodb
+	@sudo tar -xf dynamodb_local_latest.tar.gz -C /usr/local/dynamodb
+	@rm dynamodb_local_latest.tar.gz
```
tjholm, line 30: Does this need to install the firestore emulator as well?

My personal preference with integration tests would be to somehow containerise them, to simplify setup and separate them from our unit tests as well.

Without mocked unit testing it becomes hard to test plugin behaviour against unexpected errors from upstream services. I think for our unit tests we should implement similar mocking that we have in our go-sdk, and assert only on interactions between the plugins and their upstream services (e.g. correct translation of queries and construction of requests). Our method of emulating underlying behaviour from the mocks before was incorrect we should be asserting on method spys only, which we could auto generate with something like go-mock.

plugins/document/document.go
```diff
@@ -0,0 +1,219 @@
...
+func Stack() (*utils.NitricStack, error) {
+	if stack == nil {
+		nitricStack, err := utils.NewStackDefault()
+		if err != nil {
+			return nil, fmt.Errorf("error loading Nitric stack definition: %v", err)
+		}
+		stack = nitricStack
+	}
+	return stack, nil
+}
```
tjholm, line 54: As discussed we won't need the definition for now, should probably be in a different package as well.

plugins/document/document.go
```diff
@@ -0,0 +1,219 @@
...
+	for index, exp := range sortedExps {
+		if index < (len(sortedExps) - 1) {
+			nextExp := sortedExps[index+1]
+
+			if exp.Operand == nextExp.Operand &&
+				((exp.Operator == ">" && nextExp.Operator == "<") ||
+					(exp.Operator == ">" && nextExp.Operator == "<=") ||
+					(exp.Operator == ">=" && nextExp.Operator == "<")) {
+
+				return fmt.Errorf("range expression not supported with DynamoDB (use operators >= and <=) : %v", exp)
```
tjholm, line 213: As we're accepting this as part of our unified API, is it necessary to mention DynamoDB here? Could be confusing for users deployed to other clouds seeing this error.

plugins/document/document_suite_test.go
```diff
@@ -21,7 +21,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func TestCognito(t *testing.T) {
+func TestKV(t *testing.T) {
```
tjholm, line 24: TestDocument?

plugins/document/firestore/firestore_test.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"syscall"
+
+	"cloud.google.com/go/firestore"
+	ds_plugin "github.com/nitric-dev/membrane/plugins/document/firestore"
+	"github.com/nitric-dev/membrane/plugins/document/test"
```
tjholm, line 26: I don't have a better suggestion on where to put this module right now, but placing it in parallel with the plugin implementations feels a bit odd.

utils/stack_test.go
```diff
@@ -0,0 +1,111 @@
...
+				os.Setenv(utils.NITRIC_HOME, "test/")
+				os.Setenv(utils.NITRIC_YAML, "nitric-valid.yaml")
+				stack, err := utils.NewStackDefault()
+				Expect(stack).ToNot(BeNil())
+				Expect(err).To(BeNil())
+				Expect(stack.Collections).To(HaveLen(2))
+			})
+		})
+	})
+})
```
tjholm, line 111: If we can move this and related work to a separate branch that would be good.

adapters/grpc/document_grpc.go
```diff
@@ -0,0 +1,166 @@
...
+	documentPlugin sdk.DocumentService
+}
+
+func (s *DocumentServer) checkPluginRegistered() (bool, error) {
+	if s.documentPlugin == nil {
+		return false, status.Errorf(codes.Unimplemented, "Document plugin not registered")
+	}
+
+	return true, nil
+}
```
tjholm, line 41: I know this is a copy paste carry over from other plugins, in hindsight the boolean return is very redundant, can get rid of it if you'd like. No need to change other plugins though that can be done separately.

tjholm (changes requested):

sdk/document.go
```diff
@@ -0,0 +1,67 @@
...
+	PagingToken map[string]string
+}
+
+// The base Document Plugin interface
+// Use this over proto definitions to remove dependency on protobuf in the plugin internally
+// and open options to adding additional non-grpc interfaces
+type DocumentService interface {
+	Get(Key, *Key) (map[string]interface{}, error)
+	Set(Key, *Key, map[string]interface{}) error
+	Delete(Key, *Key) error
```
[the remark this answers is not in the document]
medgar-nitric: Yes the split between reference vs value is indicating optional values. There are internal key validation methods, to ensure the key contents are valid. I am happy to change this to pass by reference.

adapters/grpc/document_grpc.go
medgar-nitric: sure

plugins/document/firestore/firestore_test.go
medgar-nitric: do you mean this being an integration test, feels a bit odd to have this here?

plugins/document/document_suite_test.go
medgar-nitric: Thanks

plugins/document/document.go
medgar-nitric: Yep, makes sense.

plugins/document/dynamodb/dynamo_db_test.go
```diff
@@ -0,0 +1,148 @@
...
+		"-inMemory",
+	}
+	cmd := exec.Command("/usr/bin/java", args[:]...)
+	if err := cmd.Start(); err != nil {
+		panic(fmt.Sprintf("Error starting Local DynamoDB %v : %v", cmd, err))
+	}
+	fmt.Printf("Started Local DynamoDB (PID %v) and loading data...\n", cmd.Process.Pid)
+
+	return cmd
+}
```
tjholm, line 46: I would like to discuss our testing strategy for integrations vs. unit tests. These tests are great, but I would like to separate them out into their own suite and make target.

Something like:
```makefile
# Pure unit tests
test-unit:
  ... run unit tests
 
# Integration tests
test-integration:
  ... run integration tests (this includes dependencies on external software)

# Aggregate test make target.
test: test-unit, test-integration
 @echo running tests
```
Also it would be ideal to containerise the local integration test suite and provide a either a pre-built image or an in repo dockerfile that is prepared inline to simplify the process. Running the tests is one of the first things potential contributors will probably do and lowering potential for failure and simplifying setup would help us lower rejection.

It could also be handy to add a timeout for starting the external processes as well. Currently the test suite just hangs, on a machine without the required files.

tjholm (approved):

PR #83, opened 2021-07-14
wip: support recursive document keys

jyecusch:
Added validation for max depth


sdk/document.go
```diff
@@ -16,13 +16,19 @@ package sdk
...
+const MaxSubCollectionDepth int = 1
+
 type Collection struct {
 	Name   string
 	Parent *Key
 }
 
 type Key struct {
-	Collection Collection
+	Collection *Collection
```
medgar-nitric, line 31: Not sure why changing this to be pointer?  Is this to be consistent with Parent being pointer with Collection. From memory perspective its probably more compact as is, but this probably trivial in broader context.

jyecusch: It was for consistency with how the Parent is a pointer in the collection, the other is because the method that converts a collection from the pb struct was also returning a pointer. Probably not a big deal either way, like you said.

adapters/grpc/document_grpc.go
```diff
@@ -110,7 +124,27 @@ func NewDocumentServer(docPlugin sdk.DocumentService) pb.DocumentServiceServer {
 	}
 }
 
-func toPbDoc(doc *sdk.Document) (*pb.Document, error) {
+// isInvalidSubCollectionDepth - returns an error if the provided collection exceeds the maximum supported
```
medgar-nitric, line 127: Good addition. Currently most of the validation parameter validation code is in the file `/plugins/documents/document.go`.  This is used across all the document services.  I think the parameter validation code should be centralized.  

Next question is whether we perform validation in the adaptor layer.  The advantages of this is we protect against the plugin implementations which forget to do this. However is is probably only good for relatively simple validations, like what we currently do.  

Having the plugin services also enforce their contracts isn't terrible the services self contained and we ensure their behavior is consistent with the service tests.

Currently there there is the possibility of a NPE with the collection parameter, being null.

medgar-nitric: Also good if we develop a pattern for across all our services. Nice error message by the way :)

jyecusch: Moved the code as suggested, agree it's better to keep it together and it would be worth developing a pattern.

medgar-nitric (approved):

tjholm (approved):

PR #84, opened 2021-07-14
Service suffix update

tjholm:
NOTE: This will need to be closely co-ordinated with sdk updates.


pluggable_membrane.go
```diff
@@ -74,7 +74,7 @@ func main() {
 	var eventingService sdk.EventService = nil
 	var gatewayService sdk.GatewayService = nil
 	var keyValueService sdk.KeyValueService = nil
-	var queueService sdk.QueueService = nil
+	var QueueService sdk.QueueService = nil
```
medgar-nitric, line 77: Is there a reason for leading uppercase character ?

tjholm: find and replace error from a rename

medgar-nitric (approved):

jyecusch (approved):

PR #85, opened 2021-07-14
chore: improve client side plugin service error reporting

medgar-nitric:
This change is to improve client side error reporting from plugin services:

Previously:
```
Error occurred handling request GET / with function: com.example.Handler
io.grpc.StatusRuntimeException: UNKNOWN: provide non-nil key
...
```

PR changes to:
```
Error occurred handling request GET / with function: com.example.Handler
io.grpc.StatusRuntimeException: INVALID_ARGUMENT: DocumentServer.Set: provide non-nil key
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,7 +248,7 @@ A Nitric SDK service invocation sequence diagram is provided below.
 
 1. SDK Client - Application process makes SDK service call via GRPC client
 2. GRPC Server - Membrane process GRPC server receives call, and registered service handles call [`/pkg/membrane`]
-3. GRPC Document Service - delegates the call to the service adaptor [`/interfaces/nitric/v1/doucment_grpc.pb.go`] 
+3. GRPC Document Service - delegates the call to the service adaptor [`/interfaces/nitric/v1/document_grpc.pb.go`] 
 4. GRPC Document Adaptor - service adaptor delegates call to Cloud service plugin [`/pkg/adapters/grpc/document_grpc.go`] 
 5. Document Service Plugin (DynamoDB) - Cloud service plugin makes remote call to Cloud's API [`/pkg/plugins/document/dynamodb/dynamodb.go`]
 6. DynamoDB API - Cloud API makes remote call to Cloud service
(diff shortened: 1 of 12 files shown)


tjholm (dismissed):
LGTM.

Eventually I think we should remove the internal sdk package, and place each of the plugin interfaces in their respective plugin directories. So in future we'll have to find somwhere else for the sdk error methods to live.

medgar-nitric:
I was thinking about whether the sdk package was the right place for this IllegalArgumentError type, I think we could put this is a plugins package instead. Which is in the root of the plugins structure:

```
"github.com/nitric-dev/membrane/plugins"
```

README.md
```diff
@@ -242,8 +247,8 @@ A Nitric SDK service invocation sequence diagram is provided below.
 </p>
 
 1. SDK Client - Application process makes SDK service call via GRPC client
-2. gRPC Server - Membrane process gRPC server receives call, and registered service handles call [`/membrane`]
-3. gRPC Service KV - server delegates the call to the service adaptor [`/interfaces/nitric/v1`] 
-4. KV Adaptor (gRPC) - service adaptor delegates call to Cloud service plugin [`/adapters/grpc`] 
-5. KV Plugin (DynamoDB) - Cloud service plugin makes remote call to Cloud's API [`/plugins/...`]
+2. GRPC Server - Membrane process GRPC server receives call, and registered service handles call [`/pkg/membrane`]
+3. GRPC Document Service - delegates the call to the service adaptor [`/interfaces/nitric/v1/doucment_grpc.pb.go`]
```
tjholm, line 251: doucment -> document

pluggable_membrane.go
```diff
@@ -16,14 +16,15 @@ package main
...
-	membrane2 "github.com/nitric-dev/membrane/pkg/membrane"
-	utils2 "github.com/nitric-dev/membrane/pkg/utils"
 	"log"
 	"os"
 	"plugin"
 	"strconv"
 	"strings"
 
+	membrane2 "github.com/nitric-dev/membrane/pkg/membrane"
+	utils2 "github.com/nitric-dev/membrane/pkg/utils"
```
tjholm, line 26: the `<packagename>2` imports are hangovers from the pkg refactor, that was done earlier, would you be able to remove these and update references in this file?

tjholm (changes requested):
Just a typo and one other package name fix from the file move before other than that looks good.

tjholm (approved):

PR #90, opened 2021-07-15
fix(boltdb): errors in unstructured data queries and non-strings

jyecusch:
Querying for a property that isn't present on all documents would cause the queries to fail.
Also, querying for non-string values would fail.


tjholm (approved):

medgar-nitric:
We should be adding a test case to the integration for this.

PR #95, opened 2021-07-19
feat: added doc merge support

medgar-nitric:
┆Issue is synchronized with this [Clickup task](https://app.clickup.com/t/1dyn9cq) by [Unito](https://www.unito.io)

### DIFF
--- a/pkg/plugins/document/boltdb/boltdb.go
+++ b/pkg/plugins/document/boltdb/boltdb.go
@@ -16,12 +16,13 @@ package boltdb_service
 
 import (
 	"fmt"
-	"github.com/nitric-dev/membrane/pkg/plugins/document"
-	"github.com/nitric-dev/membrane/pkg/utils"
 	"os"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/nitric-dev/membrane/pkg/plugins/document"
+	"github.com/nitric-dev/membrane/pkg/utils"
 
 	"github.com/Knetic/govaluate"
 	"github.com/asdine/storm"
@@ -76,7 +77,7 @@ func (s *BoltDocService) Get(key *sdk.Key) (*sdk.Document, error) {
 	return toSdkDoc(doc), nil
 }
 
-func (s *BoltDocService) Set(key *sdk.Key, content map[string]interface{}) error {
+func (s *BoltDocService) Set(key *sdk.Key, content map[string]interface{}, merge bool) error {
 	err := document.ValidateKey(key)
 	if err != nil {
 		return err
@@ -93,7 +94,24 @@ func (s *BoltDocService) Set(key *sdk.Key, content map[string]interface{}) error
 	defer db.Close()
 
 	doc := createDoc(key)
-	doc.Value = content
+
+	if merge {
+		if len(content) == 0 {
+			return nil
+		}
+
+		err = db.One(idName, doc.Id, &doc)
+		if err != nil {
+			// If doc not found save will create a new record
+			doc.Value = make(map[string]interface{})
+		}
+		for name, value := range content {
+			doc.Value[name] = value
+		}
+
+	} else {
+		doc.Value = content
+	}
 
 	return db.Save(&doc)
 }
--- a/pkg/plugins/document/boltdb/boltdb.go
+++ b/pkg/plugins/document/boltdb/boltdb.go
@@ -21,6 +21,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/imdario/mergo"
+
 	"github.com/nitric-dev/membrane/pkg/plugins/document"
 	"github.com/nitric-dev/membrane/pkg/utils"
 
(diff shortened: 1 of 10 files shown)


pkg/plugins/document/boltdb/boltdb.go
```diff
@@ -93,7 +94,24 @@ func (s *BoltDocService) Set(key *sdk.Key, content map[string]interface{}) error
...
+		}
+
+		err = db.One(idName, doc.Id, &doc)
+		if err != nil {
+			// If doc not found save will create a new record
+			doc.Value = make(map[string]interface{})
+		}
+		for name, value := range content {
+			doc.Value[name] = value
+		}
```
tjholm, line 110: This probably needs more discussion as to whether we want to merge to be deep or shallow.

medgar-nitric: I presume you mean if the value is a ```map[string]interface{}``` ?

medgar-nitric: Have done some testing, Firestore out of the box performs a deep merge of maps. However for arrays it will replace them. 
While there is some support for Array Union operations, it is not supported in the Go API: 
https://firebase.google.com/docs/firestore/manage-data/add-data#go_11

DynamoDB does not support merging out of the box. 

At this stage I think we should write our own merge function, so we have consistent behavior across the database services.
This library looks good; https://github.com/imdario/mergo

medgar-nitric: Mergo library works well but for array elements it just does a replace.  I think this is OK so long as we document this behavior. I don't think there is any standard way to merge an arrays, e.g. append? what about ordering? duplication of elements?

tjholm: Just thinking more on this, are there any mutex concerns with the read + replace, or can we perform these operations in a transaction for databases that don't natively support merging?

jyecusch: This is the only thing I have a concern about, a sequence like this:

1. Process A reads doc, ready to perform a merge
2. Process B writes to doc
3. Process A writes merged, but out of date doc, blowing away changes from Process B

As long as it's consistent between providers it probably doesn't matter too much, but I'd be concern the Firestore implementation might behave differently to the others

medgar-nitric: I think the only way we are going to resolve this is some form of conditional update. Note I think developers would still have the same issue here doing this manually, but it would be their responsibility/fault.  Some further investigation sounds appropriate.

medgar-nitric: Looking at the firestore documentation the set operation is atomic, which makes sense.  

With dynamodb I have implemented a conditional update when performing a merge operation.  This stores an additional _hash attribute in the document which is stripped out of the returned content like the _pk, _sk values.  For the hashing algorithm I used CRC64 because I am concerned about the additional storage size of a larger hash.  You can see an example below to give you an idea.  Interested in your take on this.

CRC64 (hex encoded)
----------------------
letter:A, _pk:1, _sk:items#01, _hash:e43daa65c1edfc59

SHA256 (hex encoded)
--------------------
letter:A, _pk:1, _sk:items#01, _hash:504f183dda5d0a1490de66c75e547a68df860d2366739776191cb7cc49358f04 

Using ascii85 this _hash value length can be reduced a bit more, however values start to look pretty strange with punctuation characters.

CRC64 (ascii85 encoded)
---------------------------
letter:A, _pk:1, _sk:items#01, _hash:=o@['AZ@oS

medgar-nitric: Thinking about this topic I think the greater danger in this area is not having a generalized conditional set/update operation. In this use case the UI pulls the entire document into the view and then performs an set which overwrites another operation. Our existing API doesn't really have any capability for dealing with this.

With dynamo and mongo they support conditional updates. Firestore doesn't appear to but supports conditional updates. However it supports transactional updates, so a read operation could be performed in a transaction to compare the incoming record to the existing database value before performing an update.

I think going down this path would require a bit of rework on our contracts API and SDK, as I would envisage us returning a Document type on the get operation containing metadata (keys, hash).

```
DocumentRef docRef = Documents.collection("customer").doc(key);

// Performs DB load operation (same doc object type returned in query)
Document doc = docRef.get();

Key key = doc.getKey();
String hash = doc.getHash();
Map content = doc.getContent(); 

// Update content
docRef.set(doc);
```

medgar-nitric: I would think providing some set options would be good, so people can be fine grained in how they want the operation to be performed. The options I would envisage would be:

- OVERWRITE : over writes existing document contents, least expensive operation probably use in bulk operations
- MERGE : performs firestore merge and dynamodb conditional merge operation
- CONDITIONAL: performs firestore transactional update and dynamodb conditional put operation

```
docRef.set(doc)
docRef.set(doc, SetOptions.MERGE)
docRef.set(doc, SetOptions.OVERWRITE)
docRef.set(doc, SetOptions.CONDITIONAL)
```

The default choice is interesting, CONDITIONAL would be the safest but would return errors if there was a conflict and also more expensive. OVERWRITE conversely is easiest to use and cheapest, but will clobber other writes.

PR #96, opened 2021-07-20
feat: add delete child docs

medgar-nitric:


tjholm (changes requested):
Is this change unrelated to the Doc set merge changes?

If so can you update contracts to current main?

medgar-nitric:
Not related to Doc set merge.   I rebased this with main, but now having some issue checking out contracts on build server. I will need some help resolving this.

pkg/plugins/document/dynamodb/dynamodb.go
```diff
@@ -122,17 +124,49 @@ func (s *DynamoDocService) Delete(key *sdk.Key) error {
...
+		}
+
+		err = s.processDeleteQuery(key, resp)
+		if err != nil {
+			return fmt.Errorf("error performing delete: %v", err)
+		}
+
+		lastEvaluatedKey := resp.LastEvaluatedKey
+
+		for len(lastEvaluatedKey) > 0 {
```
jyecusch, line 153: Could we reduce this duplicate delete query code using a recursive function instead? Or, alternatively put it in a while true loop, with a conditional 'break' statement once the lastEvaluatedKey is nil? Something like:

```
var lastEvaluatedKey = nil
for {
  queryInput := createDeleteQuery(key, lastEvaluatedKey)
  resp, err := s.client.Query(queryInput)
  if err != nil {
	  return fmt.Errorf("error performing delete: %v", err)
  }
  
  err = s.processDeleteQuery(key, resp)
  if err != nil {
	  return fmt.Errorf("error performing delete: %v", err)
  }
  
  lastEvaluatedKey = resp.LastEvaluatedKey
  if lastEvaluatedKey == nil {
    break
  }
}
```

pkg/plugins/document/firestore/firestore.go
```diff
@@ -83,13 +83,53 @@ func (s *FirestoreDocService) Delete(key *sdk.Key) error {
...
+			return fmt.Errorf("error deleting value: %v", err)
+		}
+
+		for {
+			iter := subColRef.Limit(100).Documents(s.context)
+			numDeleted := 0
+
+			batch := s.client.Batch()
+			for {
+				subDoc, err := iter.Next()
```
jyecusch, line 103: These lines could probably be merged into the 'for' loop for clarity on what's being looped over

```go
for subDoc, err := iter.Next(); err != iterator.Done {
```

pkg/plugins/document/firestore/firestore.go
```diff
@@ -83,13 +83,53 @@ func (s *FirestoreDocService) Delete(key *sdk.Key) error {
...
+		subColRef, err := iter.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return fmt.Errorf("error deleting value: %v", err)
+		}
+
+		for {
+			iter := subColRef.Limit(100).Documents(s.context)
```
jyecusch, line 98: Could we rename this inner iterator variable to avoid confusion with the collections iterator on line 87?

jyecusch (changes requested):

pkg/plugins/document/dynamodb/dynamodb.go
medgar-nitric: Sure will take a look

pkg/plugins/document/firestore/firestore.go
medgar-nitric: Good point, thx

pkg/plugins/document/firestore/firestore.go
medgar-nitric: Not sure the alternative is much better:
```
subDoc, err := docsIterator.Next()
for ; err != iterator.Done; subDoc, err = docsIterator.Next() {
    if err != nil {
        return err
    }

    batch.Delete(subDoc.Ref)
    numDeleted++
}
```

pkg/plugins/document/dynamodb/dynamodb.go
medgar-nitric: Have done.

pkg/plugins/document/firestore/firestore.go
medgar-nitric: Have done

pkg/plugins/document/firestore/firestore.go
```diff
@@ -82,13 +82,53 @@ func (s *FirestoreDocService) Delete(key *document.Key) error {
...
+			// If no more to delete, completed
+			if numDeleted == 0 {
+				break
+			}
+
+			_, err := batch.Commit(s.context)
+			if err != nil {
+				return err
+			}
+		}
```
tjholm, line 123: The second inner loop here might be redundant, let me know if I'm missing something but it seems equivalent to:
```go
	collsIterator := doc.Collections(s.context)
	for {
		subColRef, err := collsIterator.Next()
		if err == iterator.Done {
			break
		}
		if err != nil {
			return fmt.Errorf("error deleting value: %v", err)
		}

		docsIterator := subColRef.Limit(100).Documents(s.context)
		numDeleted := 0

		batch := s.client.Batch()
		for {
			subDoc, err := docsIterator.Next()
			if err == iterator.Done {
				break
			}
			if err != nil {
				return err
			}

			batch.Delete(subDoc.Ref)
			numDeleted++
		}

		// If no more to delete, completed
		if numDeleted == 0 {
		        // replace break with continue to restart the outer loop
			continue
		}

		_, err := batch.Commit(s.context)
		if err != nil {
			return err
		}
	}
```
Assuming the break was to skip an empty commit.

tjholm (changes requested):

pkg/plugins/document/firestore/firestore.go
medgar-nitric: The reason for the additional loop is there is a batch limit on the number of items which can be deleted (maximum is 500). I have updated this block of code to make it more explicit what it is doing and increased the batching size.

This code was derived from GCP doco examples: https://github.com/GoogleCloudPlatform/golang-samples/blob/09368dd82c1ebc3a8c43e258abc6c1fd9e33277f/firestore/firestore_snippets/save.go#L271-L308

pkg/plugins/document/firestore/firestore.go
tjholm: you could include init in the for loop:

```go
for subDoc, err := docsIterator.Next(); err != iterator.Done; subDoc, err = docsIterator.Next() {
    if err != nil {
        return err
    }

    batch.Delete(subDoc.Ref)
    numDeleted++
}
```

pkg/plugins/document/firestore/firestore.go
tjholm: cool, clearer with the changes, I'd still recommend including the scope in each for loop per Jye's recommendation above to make it clearer.

pkg/plugins/document/firestore/firestore.go
medgar-nitric: Done

tjholm (approved):

tjholm:
One quick question on the commits, is this a feature or a fix? Seems like missing required behaviour. This branch will need a rebase anyway due to conflicts, this could be resolved as part of a squash rebase with conventional commit rewording.

jyecusch (dismissed):

tjholm (approved):

jyecusch (approved):

PR #97, opened 2021-07-21
Remove sdk package and move base plugins into plugins dir.

tjholm:
A basic refactor with this one, that removes the central SDK package, and moves each of the plugin definitions into their plugin packages.


pkg/adapters/grpc/event_grpc_test.go
```diff
@@ -17,24 +17,25 @@ package grpc_test
 import (
 	"context"
 	"fmt"
+
 	"github.com/nitric-dev/membrane/pkg/adapters/grpc"
 
 	v1 "github.com/nitric-dev/membrane/interfaces/nitric/v1"
-	"github.com/nitric-dev/membrane/pkg/sdk"
+	"github.com/nitric-dev/membrane/pkg/plugins/eventing"
```
medgar-nitric, line 24: If we are doing this change, I think it would be a good time to consider changing this package name to events or event:

```github.com/nitric-dev/membrane/pkg/plugins/events```

File: pkg/plugins/eventing/sns/sns.go [hunk 2]
```diff
@@ -17,18 +17,19 @@ package sns_service
 import (
```
medgar-nitric, line 17: I am wondering whether it would be useful to add an addition level of hierarchy here to support different plugin implementations from the same vendor. 

Not sure if perfect example but you get the picture:
```
eventing/aws/sns/sns.go
eventing/aws/sqs/sqs.go
```

pkg/plugins/eventing/event.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 Nitric Pty Ltd.
+//
```
medgar-nitric, line 2: Would recommend moving this type into the file: plugins.go

pkg/plugins/queue/task.go
```diff
@@ -12,13 +12,12 @@
 // See the License for the specific language governing permissions and
```
medgar-nitric, line 12: Would recommend moving these types into the ```plugin.go``` file.

medgar-nitric:
I like the refactor it makes the code base more modular.

pkg/adapters/grpc/event_grpc_test.go
tjholm: agreed, will pop that in as well

pkg/plugins/eventing/sns/sns.go
tjholm: providers are intentionally left out of the hierarchy, they were originally there but were more pain than it was worth.

With the current hierarchy `eventing/sns/sns.go` and `eventing/sqs/sqs.go` also work without adding the extra step. Theres always the option of including the base provider to create a fully qualified product name i.e. `aws_sns`.

pkg/plugins/eventing/event.go
tjholm: go files are always package scoped, I'm curious as to what advantage moving it would provide?

pkg/plugins/queue/task.go
tjholm: see above

pkg/adapters/grpc/event_grpc_test.go
tjholm: done

pkg/plugins/eventing/event.go
medgar-nitric: Idea is to develop some consistency around this with the various plugins. Do we put dependent type in with the services ore keep them in a separate file.

medgar-nitric (dismissed):
Happy with refactor. 

I would like us to develop some pattern around file names for the plugins with regard to services, types, etc.  While there aren't strict opinions around this in go, I think it is still useful for us to develop some approach.

jyecusch (approved):

medgar-nitric (approved):

PR #103, opened 2021-07-28
Base Secret Service Plugin

tjholm:
Base plugin interface and SDK that secret plugins will be built on. As well as adapters for gRPC secret service.

### DIFF
--- a/pkg/adapters/grpc/secret_grpc.go
+++ b/pkg/adapters/grpc/secret_grpc.go
@@ -51,7 +51,7 @@ func (s *SecretServer) Put(ctx context.Context, req *pb.SecretPutRequest) (*pb.S
 				},
 			}, nil
 		} else {
-			return nil, err
+			return nil, NewGrpcError("SecretService.Put", err)
 		}
 	} else {
 		return nil, err
@@ -76,7 +76,7 @@ func (s *SecretServer) Access(ctx context.Context, req *pb.SecretAccessRequest) 
 				Value: s.Value,
 			}, nil
 		} else {
-			return nil, err
+			return nil, NewGrpcError("SecretService.Access", err)
 		}
 	} else {
 		return nil, err
--- a/pkg/adapters/grpc/secret_grpc.go
+++ b/pkg/adapters/grpc/secret_grpc.go
@@ -1,3 +1,17 @@
+// Copyright 2021 Nitric Pty Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package grpc
 
 import (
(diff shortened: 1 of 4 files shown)


tjholm:
Plugin PRs will be raised once this is in.

davemooreuws (approved):
LGTM

pkg/plugins/secret/types.go
```diff
@@ -0,0 +1,37 @@
...
+	Name string
+}
+
+// SecretVersion - A version of a secret
+type SecretVersion struct {
+	Secret  *Secret
+	Version string
+}
+
+// SecretAccessResponse - Return value for a secret access request
```
medgar-nitric, line 28: Can we add a comment that using the Version value of "latest" will return the latest version of the secret.

pkg/adapters/grpc/secret_grpc.go
```diff
@@ -0,0 +1,90 @@
...
+	"google.golang.org/grpc/status"
+)
+
+// GRPC Interface for registered Nitric Secret Plugins
+type SecretServer struct {
+	pb.UnimplementedSecretServiceServer
+	secretPlugin secret.SecretService
+}
+
+func (s *SecretServer) checkPluginRegistered() error {
```
medgar-nitric, line 32: Note I removed service plugin registered check with other plugins as recommended.

pkg/adapters/grpc/secret_grpc.go
```diff
@@ -0,0 +1,90 @@
...
+			return &pb.SecretPutResponse{
+				SecretVersion: &pb.SecretVersion{
+					Secret: &pb.Secret{
+						Name: r.SecretVersion.Secret.Name,
+					},
+					Version: r.SecretVersion.Version,
+				},
+			}, nil
+		} else {
+			return nil, err
```
medgar-nitric, line 54: Please report errors as:
```
return nil, NewGrpcError("SecretService.Put", err)
```

pkg/adapters/grpc/secret_grpc.go
```diff
@@ -0,0 +1,90 @@
...
+				SecretVersion: &pb.SecretVersion{
+					Secret: &pb.Secret{
+						Name: s.SecretVersion.Secret.Name,
+					},
+					Version: s.SecretVersion.Version,
+				},
+				Value: s.Value,
+			}, nil
+		} else {
+			return nil, err
```
medgar-nitric, line 79: Please report errors as:
```
return nil, NewGrpcError("SecretService.Access", err)
```

medgar-nitric (changes requested):

pkg/adapters/grpc/secret_grpc.go
tjholm: That one snuck past me, when was this recommended? If the plugins are nil in these cases it will cause the membrane to crash entirely, as the constructor methods aren't providing safety here either. If we're going to remove these checks I would prefer the safety be implemented in constructor methods and RPC server registration skipped an runtime in the membrane.

pkg/adapters/grpc/secret_grpc.go
tjholm: done

pkg/adapters/grpc/secret_grpc.go
tjholm: done

pkg/plugins/secret/types.go
tjholm: done

pkg/adapters/grpc/secret_grpc.go
medgar-nitric: It was something you recommended during a documents PR.  Simple to put these back.

medgar-nitric (approved):

pkg/adapters/grpc/secret_grpc.go
tjholm: I remember recommending removing the `bool` return on the method (originally it returned ok like a cast), but not the entire method.
https://github.com/nitrictech/membrane/pull/76#discussion_r665080267

jyecusch (approved):

PR #106, opened 2021-07-29
feat: Add lease id to dev queue tasks

HomelessDinosaur:


jyecusch (approved):
I'm happy with this, for now, I think in future we might need more realistic lease functionality (e.g. releasing tasks back to the queue, etc.) since people will likely want to test that behaviour.

tjholm (approved):

PR #107, opened 2021-07-30
Add gcp secret manager plugin.

tjholm:


pkg/plugins/secret/secret_manager/secret_manager.go
```diff
@@ -0,0 +1,212 @@
...
+	"google.golang.org/grpc/status"
+)
+
+type SecretManagerClient interface {
+	AccessSecretVersion(context.Context, *secretmanagerpb.AccessSecretVersionRequest, ...gax.CallOption) (*secretmanagerpb.AccessSecretVersionResponse, error)
+	AddSecretVersion(context.Context, *secretmanagerpb.AddSecretVersionRequest, ...gax.CallOption) (*secretmanagerpb.SecretVersion, error)
+	CreateSecret(context.Context, *secretmanagerpb.CreateSecretRequest, ...gax.CallOption) (*secretmanagerpb.Secret, error)
+	GetSecret(context.Context, *secretmanagerpb.GetSecretRequest, ...gax.CallOption) (*secretmanagerpb.Secret, error)
+	UpdateSecret(context.Context, *secretmanagerpb.UpdateSecretRequest, ...gax.CallOption) (*secretmanagerpb.Secret, error)
+}
```
jyecusch, line 38: I'm assuming this is an interface for the GCP API since they don't provide one, it would be good to provide a comment if that's the case.

jyecusch (approved):

pkg/plugins/secret/secret_manager/secret_manager.go
```diff
@@ -0,0 +1,212 @@
...
+
+type secretManagerSecretService struct {
+	secret.UnimplementedSecretPlugin
+	client    SecretManagerClient
+	projectId string
+}
+
+func validateNewSecret(sec *secret.Secret, val []byte) error {
+	if sec == nil {
+		return plugins.NewInvalidArgError("secret must not be nil")
```
medgar-nitric, line 48: Current error message convention we have is: `"provide non-nil secret"`

pkg/plugins/secret/secret_manager/secret_manager.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	if len(sec.Name) == 0 {
+		return plugins.NewInvalidArgError("provide non-blank secret name")
+	}
+	if len(val) == 0 {
```
medgar-nitric, line 53: We should probably add some common secret.Name validation around what a valid name is, e.g. text, numbers, dashes and . only for all the plugins.  Worried about / character in Name with version parsing later.

pkg/plugins/secret/secret_manager/secret_manager_test.go
```diff
@@ -0,0 +1,310 @@
...
+
+				It("Should return an error", func() {
+					var emptySecretName = &secret.Secret{}
+					_, err := secretPlugin.Put(emptySecretName, testSecretVal)
+
+					Expect(err).Should(HaveOccurred())
+					Expect(err.Error()).To(Equal("provide non-blank secret name"))
+				})
+			})
+
```
medgar-nitric, line 177: Could we add some dodgy secret name validation as well.

medgar-nitric (changes requested):

pkg/plugins/secret/secret_manager/secret_manager_test.go
tjholm: new issue for this, as we should add this in the adapter layer.

pkg/plugins/secret/secret_manager/secret_manager.go
tjholm: will raise a new issue for this, as it will be common to all plugins (adapter layer validation).

pkg/plugins/secret/secret_manager/secret_manager.go
tjholm: done

pkg/plugins/secret/secret_manager/secret_manager.go
tjholm: done

medgar-nitric (approved):

jyecusch (approved):

PR #109, opened 2021-07-30
Add local dev secret plugin

tjholm:


pkg/plugins/secret/dev/dev.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	if len(val) == 0 {
+		return nil, fmt.Errorf("provide non-blank secret value")
+	}
+
+	var versionId = uuid.New().String()
+
+	//Creates a new file in the form:
+	// DIR/Name_Version.txt
+	file, err := os.Create(s.secDir + "/" + sec.Name + "_" + versionId + ".txt")
```
jyecusch, line 57: Should the name be validated or hashed to ensure it's a safe string to use in a filename?

pkg/plugins/secret/dev/dev.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (s *DevSecretService) Access(sv *secret.SecretVersion) (*secret.SecretAccessResponse, error) {
+	if sv.Secret.Name == "" {
+		return nil, fmt.Errorf("provide non-blank name")
+	}
+	if sv.Version == "" {
+		return nil, fmt.Errorf("provide non-blank version")
+	}
+
+	content, err := ioutil.ReadFile(s.secDir + "/" + sv.Secret.Name + "_" + sv.Version + ".txt")
```
jyecusch, line 82: Suggest moving the file name creation to a function so it can be used for both access and put

pkg/plugins/secret/dev/dev_test.go
```diff
@@ -0,0 +1,130 @@
...
+					Secret:  &secret.Secret{},
+					Version: "test-version-id",
+				})
+				By("Returning an error")
+				Expect(err).Should(HaveOccurred())
+				By("Returning a nil response")
+				Expect(response).Should(BeNil())
+			})
+		})
+		When("Getting a secret with an empty version id", func() {
```
jyecusch, line 115: Need a method and test for returning the 'latest' secret version

jyecusch (changes requested):

pkg/plugins/secret/dev/dev.go
medgar-nitric: Would be good point for us to create a shared valid resource name function we can use across many of our services.

pkg/plugins/secret/dev/dev_test.go
```diff
@@ -0,0 +1,148 @@
...
+				emptySecretName := &secret.Secret{}
+				response, err := secretPlugin.Put(emptySecretName, testSecretVal)
+				By("Returning an error")
+				Expect(err).Should(HaveOccurred())
+
+				By("Returning a nil response")
+				Expect(response).Should(BeNil())
+			})
+		})
+	})
```
medgar-nitric, line 73: Add invalid secret name.

medgar-nitric (changes requested):

pkg/plugins/secret/dev/dev.go
tjholm: have raised: https://github.com/nitrictech/membrane/issues/114

To address secret name validation.

pkg/plugins/secret/dev/dev_test.go
tjholm: Is this not already at line 87?

pkg/plugins/secret/dev/dev.go
tjholm: done

pkg/plugins/secret/dev/dev_test.go
tjholm: Have raised: https://github.com/nitrictech/membrane/issues/114 to address this in common code.

tjholm:
https://github.com/nitrictech/membrane/issues/114

Has been raised to address secret name validation, in common code. Will be done as a separate PR.

medgar-nitric (approved):

jyecusch (approved):

PR #116, opened 2021-08-03
Secret GRPC Validation

tjholm:
Adapter layer validation for the secret service.

Will add additional tests for entire adapter layer once mocking infrastructure from secret plugin PRs is in.

### DIFF
--- a/pkg/plugins/secret/utils.go
+++ b/pkg/plugins/secret/utils.go
@@ -1,0 +1,25 @@
+package secret
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/nitric-dev/membrane/pkg/plugins"
+)
+
+const rx = "^\\w+(-\\w+)*$"
+
+// ValidateSecretName - Validates a secret name
+func ValidateSecretName(secName string) error {
+	if len(secName) == 0 {
+		return plugins.NewInvalidArgError("Secret name must not be blank")
+	}
+
+	match, _ := regexp.MatchString(rx, secName)
+
+	if !match {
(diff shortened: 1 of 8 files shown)


pkg/plugins/secret/utils.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/nitric-dev/membrane/pkg/plugins"
+)
+
+const rx = "^\\w+(-\\w+)*$"
+
+// ValidateSecretName - Validates a secret name
+func ValidateSecretName(secName string) error {
+	if len(secName) == 0 {
+		return plugins.NewInvalidArgError("Secret name must not be blank")
+	}
```
medgar-nitric, line 30: Recommend error message:  `provide non-blank secret name`

pkg/plugins/secret/utils.go
```diff
@@ -0,0 +1,39 @@
...
+
+	match, _ := regexp.MatchString(rx, secName)
+
+	if !match {
+		return plugins.NewInvalidArgError(fmt.Sprintf("Secret name must match pattern: %s", rx))
```
medgar-nitric, line 35: Recommend error message:  `provide secret name matching pattern: %s`

medgar-nitric (changes requested):

pkg/plugins/secret/utils.go
tjholm: done

pkg/plugins/secret/utils.go
tjholm: done

jyecusch (approved):

medgar-nitric (approved):

PR #121, opened 2021-08-03
fix: retrieve deployed collection names in AWS

jyecusch:
More work required in future, but tracked in a separate issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,12 @@ The Membrane is at the heart of the solution. Nitric applications communicate wi
   - Queues
   - Key/Value Store
   - Storage & Buckets
+  - Document Store
+  - Secret Store
 
 Additional services on our roadmap include:
 
   - RDS
-  - Document Store
   - Configuration
   - Logging
   - Authentication
--- a/README.md
+++ b/README.md
@@ -16,7 +16,6 @@ The Membrane is at the heart of the solution. Nitric applications communicate wi
 
   - Events
   - Queues
-  - Key/Value Store
   - Storage & Buckets
   - Document Store
   - Secret Store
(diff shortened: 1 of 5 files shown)


jyecusch:
depends on https://github.com/nitrictech/cli/pull/129

pkg/plugins/document/dynamodb/dynamodb.go
```diff
@@ -556,20 +580,37 @@ func isBetweenEnd(index int, exps []document.QueryExpression) bool {
...
 	coll := collection
 	for coll.Parent != nil {
 		coll = *coll.Parent.Collection
 	}
 
-	return aws.String(coll.Name)
+	// TODO: The following method for determining the deployment specific table name from the nitric name is unreliable.
+	//	a new design is in process and will replace this method for locating tables.
+	out, err := s.client.ListTables(&dynamodb.ListTablesInput{})
+
```
medgar-nitric, line 592: I think we should be caching these table name look results in memory. We can have a retry if the table value is not found, was not found to handle any deployment timing issues.

pkg/plugins/document/dynamodb/dynamodb.go
```diff
@@ -556,20 +580,37 @@ func isBetweenEnd(index int, exps []document.QueryExpression) bool {
...
+	if err != nil {
+		return nil, fmt.Errorf("encountered an error retrieving the bucket list: %v", err)
```
medgar-nitric, line 594: Change message to: `encountered an error retrieving the table list: %v`

medgar-nitric (changes requested):

pkg/plugins/document/dynamodb/dynamodb.go
jyecusch: Good suggestion, I've added some basic caching now

tjholm (approved):

README.md
```diff
@@ -18,11 +18,12 @@ The Membrane is at the heart of the solution. Nitric applications communicate wi
   - Queues
   - Key/Value Store
```
medgar-nitric, line 19: Should remove Key/Value Store

jyecusch: Done

medgar-nitric (approved):

PR #125, opened 2021-08-04
Fix: Http Gateways not shutting down correctly

tjholm:
Have also included refactoring increase maintainability of http base gateway plugins by allowing them to share a common base plugin, with intercepting middleware.

Middleware is very bare bones at the moment, but we can add additional functionality and richness later.


medgar-nitric (approved):
Nice refactor.

jyecusch (approved):

PR #126, opened 2021-08-05
fix: Add generated validators for gRPC adapters.

tjholm:
Adds generated payload validation for all gRPC adapters (except for FaaS at the moment). This essentially adds deep validation to all our payloads.

Example validation error (expression needs to be cleaned up):
`"error": "3 INVALID_ARGUMENT: EventService.Publish: invalid EventPublishRequest.Topic: value does not match regex pattern \"^\\\\w+([.\\\\-]\\\\w+)*$\""`

This will allow use to universally validate based on our proto contracts and avoid drift in validation implementation in individual SDKs. See: https://github.com/nitrictech/contracts/pull/42


medgar-nitric (commented):
Looks really good.

jyecusch (approved):

medgar-nitric (approved):

PR #135, opened 2021-08-17
feat: Add initial Key Vault plugin

HomelessDinosaur:


tjholm (approved):
Looks great!

jyecusch (changes requested):
One small suggested change, otherwise this looks great and should be good to go.

jyecusch (approved):

PR #143, opened 2021-08-22
feat:added grpc json error messages

medgar-nitric:
Add support JSON message component to gRPC error messages. This enables SDK to parse errors and display them in a much more useful way.  Below is Java SDK example:

```
SEVERE: error handling HTTP trigger with function 'com.example.Handler' :
io.nitric.api.NitricException: {
"code" : "Internal",
"msg" : "DB Fetch error",
"cause" : "not found",
"service" : "DocumentService.Get",
"plugin" : "BoltDocService.Get",
"args" : "[key={Collection: {Name: customer, Parent: <nil>}, Id: medgar@gmail.com}]"
}
at io.nitric.api.NitricException.build(NitricException.java:157)
at io.nitric.api.document.DocumentRef.get(DocumentRef.java:84)
at com.example.Handler.handle(Handler.java:13)
at io.nitric.faas.Faas$FaasStreamObserver.onNext(Faas.java:201)
...
Caused by: io.grpc.StatusRuntimeException: INTERNAL: {"code":"Internal","msg":"DB Fetch error","cause":"not found","service":"DocumentService.Get","plugin":"BoltDocService.Get","args":"[key={Collection: {Name: customer, Parent: \u003cnil\u003e}, Id: medgar@gmail.com}]"}
...
```

┆Issue is synchronized with this [Clickup task](https://app.clickup.com/t/1dyn91f) by [Unito](https://www.unito.io)


pkg/adapters/grpc/errors.go
```diff
@@ -15,25 +15,51 @@
...
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
+type grpcError struct {
+	code    codes.Code `json:"-"`
+	Code    string     `json:"code"`
+	Msg     string     `json:"msg,omitempty"`
+	Service string     `json:"service,omitempty"`
+}
```
tjholm, line 31: Why not implement this using the standard richer error model guidance? https://grpc.io/docs/guides/error/#richer-error-model

tjholm (changes requested):

pkg/adapters/grpc/errors.go
medgar-nitric: I have to admit I didn't really look at this. Will do now.

medgar-nitric: Looks like we would define a standard protobuf error message type, and then use a protobuf builders to build the structure, and then use the underlying Status type to add this `WithDetails()`. 

Not sure how this would be expressed on the client side, possibly it would be deserialized as "metadata" in the Java GRPC library. Would need to give it a try.

pkg/plugins/queue/plugin.go
```diff
@@ -47,6 +47,10 @@ type ReceiveOptions struct {
 	Depth *uint32 `type:"int" required:"false"`
 }
 
+func (p *ReceiveOptions) String() string {
+	return fmt.Sprintf("{QueueName: %s, Depth: %v}", p.QueueName, p.Depth)
+}
+
```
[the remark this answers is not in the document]
medgar-nitric: These types String() method is used when the service plugin method parameters are being logged in `pkg/adapters/grpc/errors.go`

```
		if len(pe.Args) > 0 {
			args := make(map[string]string)
			for k, v := range pe.Args {
				args[k] = fmt.Sprintf("%v", v)
			}
			ed.Scope.Args = args
		}
```
The intent of using the String() method was to ensure we are logging the parameter values consistently and are not logging any sensitive information, such as content or secrets.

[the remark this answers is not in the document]
tjholm: Cool that makes perfect sense thanks for pointing it out it's use :+1:. I still think we can do better to improve the quality & maintainability of this code, as manually implementing every `String()` implementation for a struct we need to maintain seems small at first but will continue to grow as we add more structs and plugins.

I think a tag based approach could be more consistent and easier to maintain going forward, and will still meet any security requirements for field privacy, it also makes the purpose much clearer.

```go
type A struct {
  A string `debug:"A"`
  B string `debug:"B"`
  C int `debug:"C"`
}
```
> Note we could use json, but I don't believe it meets the requirement of ignoring fields by omission (no tag), so would not be secure.

Fields not marked with the tag would not be included in the output:
```go
v := reflect.ValueOf(s)

for i := 0; i < v.NumField(); i++ {
    // Get the field tag value
    tag := v.Type().Field(i).Tag.Get("debug")

    // Skip if tag is not defined or ignored
    if tag == "" || tag == "-" {
      continue
    }

    // Append field value to output
}
```

While this is more code than the Stringer methods individually it would be able to support all structs we implement, would have it's own set of test cases, and would greatly reduce refactoring burden if we ever wanted to change the structure of the output.

An alternative could be an established pretty printer like: https://github.com/davecgh/go-spew (I don't like the inclusion of the `unsafe` package though even though we can explicetely exclude it, also it hasn't been maintained in some time.)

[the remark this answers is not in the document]
medgar-nitric: That sound interesting, and good approach from a security point of view being much more explicit.
For nested structures we would need to walk the graph, but would be a write once utility method hopefully.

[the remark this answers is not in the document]
medgar-nitric: done

tjholm (changes requested):

pkg/adapters/grpc/errors_test.go
```diff
@@ -55,4 +68,47 @@ var _ = Describe("GRPC Errors", func() {
...
+						Value:  "2a4wijgPq0PpwJ76IjT7&lTBZ$5SGRcq",
+					},
+				}
+
+				v5 := grpc.LogArg(data)
+				Expect(v5).To(BeEquivalentTo("{Name: name, Version: 3, Value: {Type: key}}"))
+			})
+		})
+	})
 })
```
tjholm, line 114: Happy to help with this but the cases here should be more descriptive and granular.

> This is just an example, no need to use it verbatim
```go
Context("Logging Arg", func() {
  When("String", func() {
    testString := "test"
    It("Should return value", func() {
      Expect(grpc.LogArg(testString)).To(Equal("test"))
    })
  })

  When("Int", func() {
     ...
  })

  When("Struct", func() {
    ...
  })
  
  When("Nested Struct", func() {
    ...
  })
})
```

pkg/plugins/events/event_test.go
```diff
@@ -0,0 +1,39 @@
...
+				PayloadType: "payloadtype",
+				Payload: map[string]interface{}{
+					"key": "value",
+				},
+			}
+			log := grpc.LogArg(event)
+			Expect(log).To(BeEquivalentTo("{ID: id, PayloadType: payloadtype}"))
+		})
+	})
+})
```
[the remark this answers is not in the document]
medgar-nitric: Sure will do, these tests were handy to ensure correct behaviour

pkg/adapters/grpc/errors.go
```diff
@@ -15,25 +15,127 @@
...
+			if tag == "" || tag == "-" {
+				continue
+			}
+
+			if len(str) > 1 {
+				str += ", "
+			}
+
+			field := value.Field(i)
+			str += fieldType.Name + ": " + LogArg(field.Interface())
```
tjholm, line 108: should `tag` be used instead of `fieldType.Name` here?

pkg/adapters/grpc/errors.go
```diff
@@ -15,25 +15,127 @@
...
+
+	} else if value.Kind() == reflect.Map {
+		str := "{"
+
+		for k, v := range arg.(map[string]interface{}) {
+			if len(str) > 1 {
+				str += ", "
+			}
+			str += fmt.Sprintf("%v", k) + ": " + LogArg(v)
+		}
```
tjholm, line 122: might want to catch this case and ignore none `string` map key types for now, otherwise this could result in a Panic and crash the membrane.

Alternatively could run a switch over supported successful casts for certain supported key types.

```go
switch typ := arg.(type) {
case map[string]interface{}:
    handle string map
case map[int]interface{}:
   // handle int map
default:
    // handle unsupported type
}
```
> Only add [int] as an example here, no need to actually support it at the moment.

tjholm (changes requested):
two minor changes, one to address a potential application crash the other to use the provided tag value for `log` in serialization.

PR #144, opened 2021-08-25
Feature/http headers fix

raksiv:
┆Issue is synchronized with this [Clickup task](https://app.clickup.com/t/1e4c6zm) by [Unito](https://www.unito.io)


pkg/worker/faas_http_worker.go
```diff
@@ -108,21 +109,28 @@ func (h *FaasHttpWorker) HandleHttpRequest(trigger *triggers.HttpRequest) (*trig
...
 		Data:     trigger.Body,
 		MimeType: mimeType,
 		Context: &pb.TriggerRequest_Http{
 			Http: &pb.HttpTriggerContext{
 				Path:        trigger.Path,
-				Headers:     trigger.Header,
+				Headers:     headers,
 				Method:      trigger.Method,
 				QueryParams: trigger.Query,
 			},
```
tjholm, line 136: we might still want to use old headers here as well, but as we're dropping support for PHP at the moment and it's the only runtime that uses this worker it might not be necessary and we might be able to deprecate this code altogether. Need to discuss.

tjholm (commented):
Not a change request, just a thought around the HTTP FaaS worker.

pkg/worker/faas_http_worker.go
raksiv: I'll create an item in the backlog to address this item so that we can close out this PR and move forward.

tjholm:
@raksiv these changes are being included as part of the contract merging into the membrane repo so closing this PR.

PR #157, opened 2021-09-06
Azure Storage & Queue Plugins

tjholm:
Queue and storage plugins combined, as they share underlying infrastructure (azure storage accounts).

Closes #157


pkg/plugins/storage/azblob/azblob.go
```diff
@@ -0,0 +1,185 @@
...
+
+	spt, err := azureutils.GetServicePrincipalToken(azure.PublicCloud.ResourceIdentifiers.Storage)
+	if err != nil {
+		return nil, err
+	}
+
+	cTkn := azblob.NewTokenCredential(spt.Token().AccessToken, tokenRefresherFromSpt(spt))
+
+	var accountURL *url.URL
+	if accountURL, err = url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net", storageAccountName)); err != nil {
```
raksiv, line 175: What if the user has a custom domain ? 
Would it be better to have "https://%s.blob.core.windows.net" exposed as an ENV configuration?

pkg/plugins/queue/azqueue/azqueue_test.go
```diff
@@ -0,0 +1,373 @@
...
+	"fmt"
+	//"bytes"
+	//"fmt"
+	"time"
+
+	azqueue2 "github.com/Azure/azure-storage-queue-go/azqueue"
+	"github.com/nitric-dev/membrane/pkg/plugins/queue"
+
+	//"io/ioutil"
+	//"strings"
```
raksiv, line 27: Minor tidy up to remove unwanted comments.

pkg/plugins/storage/azblob/azblob.go
```diff
@@ -0,0 +1,185 @@
...
+			codes.Internal,
+			"Unable to delete blob",
+			err,
+		)
+	}
+
+	return nil
+}
+
+const expiryBuffer = 2 * time.Minute
```
raksiv, line 139: Assume this doesn't require config? 2 mins is always adequate for the buffer?

raksiv (changes requested):

pkg/plugins/queue/azqueue/azqueue.go
```diff
@@ -0,0 +1,287 @@
...
+// Returns an adapted azqueue MessageIdUrl, which is a client for interacting with a specific message (task) in a specific queue
+func (s *AzqueueQueueService) getMessageIdUrl(queue string, messageId azqueue.MessageID) azqueueserviceiface.AzqueueMessageIdUrlIface {
+	mUrl := s.getMessagesUrl(queue)
+
+	return mUrl.NewMessageIDURL(messageId)
+}
+
+func (s *AzqueueQueueService) Send(queue string, task queue.NitricTask) error {
+	newErr := errors.ErrorsWithScope(
+		"AzqueueService.Send",
```
medgar-nitric, line 61: "AzqueueQueueService.Send"

pkg/plugins/queue/azqueue/azqueue.go
```diff
@@ -0,0 +1,287 @@
...
+
+	return tasks, nil
+}
+
+// Complete - Completes a previously popped queue item
+func (s *AzqueueQueueService) Complete(queue string, leaseId string) error {
+	newErr := errors.ErrorsWithScope(
+		"AzqueueQueueService.Complete",
+		map[string]interface{}{
+			"queue": queue,
```
medgar-nitric, line 210: add leaseId parameter

pkg/plugins/storage/azblob/azblob.go
```diff
@@ -0,0 +1,185 @@
...
+		} else {
+			tkn := spt.Token()
+			credential.SetToken(tkn.AccessToken)
+
+			return tkn.Expires().Sub(time.Now().Add(expiryBuffer))
+		}
+
+		// Mark the token as already expired
+		return time.Duration(0)
+	}
```
medgar-nitric, line 154: Is there any scope for having some common Azure token refresh code?

medgar-nitric (changes requested):

pkg/plugins/storage/azblob/azblob.go
tjholm: There is but need to examine the interface chain for the structs that implement these tokens for both azqueue and azblob to see where there is commonality, which can be done in a later PR.

pkg/plugins/queue/azqueue/azqueue.go
tjholm: done

pkg/plugins/queue/azqueue/azqueue.go
tjholm: Could leaseId be considered sensitive? This is something only the holder should know and is not necessarily common across a stack (like queue name).

pkg/plugins/queue/azqueue/azqueue_test.go
tjholm: done

pkg/plugins/storage/azblob/azblob.go
tjholm: For now, can address if it becomes a problem in future.

pkg/plugins/storage/azblob/azblob.go
tjholm: done, set for both queue and blob, and adding as part of env variables for app service deployment.

pkg/plugins/queue/azqueue/azqueue.go
tjholm: done, might be worth considering if leaseId is sensitive information though as only its holder should know this. (not common stack knowledge like queue names for example.

pkg/plugins/queue/azqueue/azqueue.go
```diff
@@ -0,0 +1,288 @@
...
+	mUrl := s.getMessagesUrl(queue)
+
+	return mUrl.NewMessageIDURL(messageId)
+}
+
+func (s *AzqueueQueueService) Send(queue string, task queue.NitricTask) error {
+	newErr := errors.ErrorsWithScope(
+		"AzqueueQueueService.Send",
+		map[string]interface{}{
+			"queue": queue,
```
medgar-nitric, line 63: Log task parameter as well.

medgar-nitric (changes requested):

pkg/plugins/queue/azqueue/azqueue.go
tjholm: done

medgar-nitric (approved):

HomelessDinosaur (approved):

davemooreuws (approved):

PR #162, opened 2021-09-09
feat: add grpc errors

medgar-nitric:


davemooreuws (approved):
Nice work

HomelessDinosaur (approved):

contracts/proto/error/v1/error.proto
```diff
@@ -0,0 +1,34 @@
...
+    // The plugin method arguments, ensure only non-sensitive data is specified.
+    map<string, string> args = 3;
+}
+
+message ErrorDetails {
+    // The developer error message, explaining the error and ideally solution.
+    string message = 1;
+
+    // The error details.
+    google.protobuf.Struct details = 2;
```
jyecusch, line 30: What's the thinking about the future use of 'details'? I see the 'cause' key being added in this PR. Would a standard set of keys be used? The reason I ask is that I feel a predefined `message` could be easier to implement on the SDK side, if the details keys were known ahead of time.

```
message ErrorDetail {
  string message = 1;
  ErrorScope scope = 2;
  string cause = 3;
  // extend as needed in future with new details attributes?
}

jyecusch: A little selfishly, I also find the support for the Struct type a little weak in some languages (e.g. python and PHP). So stronger typing can be easier to work with, in the generated code in the SDKs.

jyecusch (commented):
Overall LGTM, one question/comment about the error message contract. Besides that, probably GTG.

contracts/proto/error/v1/error.proto
medgar-nitric: Tim and I went backward and forward on the string cause vs struct details, which is the only thing used on the details.  I am happy to leave details out until we need it. We considered details as a map<string, string> however, motivation was to provide ability to deal with unknown shapes of data (list, etc).

medgar-nitric: done

tjholm (approved):

PR #168, opened 2021-09-14
Calculate the rc number for each dev release.

tjholm:
Will use the difference in merge commits between develop and main


jyecusch (approved):

davemooreuws (approved):
LGTM

PR #169, opened 2021-09-14
feat: add aws ses emails service

jyecusch:
One thing to note - we discovered that the mail API from Google Cloud is deprecated, they recommend using 3rd party services like SendGrid, Mailgun or Mailjet. Perhaps we should drop this feature using a similar recommendation. Alternatively, we can find a good method for devs to select the email service plugin they want at build time, then integrate with those 3rd parties for them with our emails service.

### DIFF
--- a/makefile
+++ b/makefile
@@ -194,5 +194,7 @@ generate-mocks:
 	@echo Generating Mock Clients
 	@mkdir -p mocks/mock_secret_manager
 	@mkdir -p mocks/secrets_manager
+	@mkdir -p mocks/aws_ses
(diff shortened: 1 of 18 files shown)


jyecusch:
No equivalent services provided by Google, Azure, etc. we'll review this again in future and suggest direct integration with AWS SES or a third-party like Mailgun for now.

medgar-nitric:
Makes sense to drop this if Azure not providing service either. Amazing GCP would deprecate a feature however.

Does make me think about DI injection done in the user's code, e.g. they develop a service which they can use with their functions, rather than doing Membrane development, which will be beyond the motivation/capability of many developers.

PR #171, opened 2021-09-16
Feature/eventing azure

HomelessDinosaur:


pkg/plugins/events/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,209 @@
...
+		for _, topic := range topicsList {
+			topics = append(topics, *topic.Name)
+		}
+		results.NextWithContext(ctx)
+	}
+
+	return topics, nil
+}
+
+func (s *EventGridEventService) GetTopicEndpoint(topicName string) (string, error) {
```
jyecusch, line 79: Can this function be private (lowercase first letter)?

pkg/plugins/events/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,209 @@
...
+			if *topic.Name == topicName {
+				return strings.TrimSuffix(strings.TrimPrefix(*topic.Endpoint, "https://"), "/api/events"), nil
+			}
+		}
+		results.Next()
+	}
+	return "", fmt.Errorf("topic with provided name could not be found")
+}
+
+func (s *EventGridEventService) NitricEventToEvent(topic string, event *events.NitricEvent) ([]eventgrid.Event, error) {
```
jyecusch, line 99: Can this be private too?

pkg/plugins/events/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,209 @@
...
+	azureEvent := []eventgrid.Event{
+		{
+			ID:          &event.ID,
+			Data:        &payload,
+			EventType:   &event.PayloadType,
+			Subject:     &topic,
+			EventTime:   &date.Time{time.Now()},
+			DataVersion: &dataVersion,
+		},
+	}
```
jyecusch, line 114: This code is always constructing a slice of events with just 1 entry. Is that intentional or should it just return the single event struct instead of a slice?

jyecusch: If it needs to be a slice, I suggest allowing a slice as input and updating the func and variable names.

pkg/plugins/events/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,209 @@
...
+			"topic": topic,
+		},
+	)
+	ctx := context.Background()
+
+	if len(topic) == 0 {
+		return newErr(
+			codes.InvalidArgument,
+			"provided invalid topic",
+			fmt.Errorf(""),
```
jyecusch, line 132: This can be `nil` instead of an empty error. Alternatively, you could provide more detail in this "cause" error message. E.g. "topic name is blank".

pkg/plugins/events/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,209 @@
...
+		)
+	}
+	if event == nil {
+		return newErr(
+			codes.InvalidArgument,
+			"provided invalid event",
+			fmt.Errorf(""),
```
jyecusch, line 139: This can be nil instead of an empty error. Alternatively, you could provide more detail in this "cause" error message. E.g. "non-nil event is required".

pkg/plugins/events/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	statusCode := fmt.Sprint(result.StatusCode)
+	if strings.Split(statusCode, "")[0] != "2" {
+		return newErr(
+			codes.Internal,
+			"returned non 200 status code",
+			fmt.Errorf(""),
+		)
+	}
```
jyecusch, line 172: I'd suggest you don't use a string here to check the status code, keep it as an int. You could also include the status in the error that's returned. E.g.

```go
if result.StatusCode < 200 || result.StatusCode >= 300 {
  return newErr(
    codes.Internal,
    "returned non 200 status code",
    fmt.Errorf(result.Status),
  )
}
```

jyecusch (changes requested):
Looking good, just a few minor changes/questions

tjholm (approved):

jyecusch (approved):

PR #172, opened 2021-09-16
boltdb document streaming

tjholm:


pkg/plugins/document/boltdb/boltdb.go
```diff
@@ -359,6 +353,61 @@ func (s *BoltDocService) Query(collection *document.Collection, expressions []do
...
+			"collection": collection,
+		},
+	)
+
+	var tmpLimit = limit
+	res, fetchErr := s.query(collection, expressions, limit, nil, newErr)
+
+	// Initial fetch
+	var documents []document.Document = res.Documents
+	var pagingToken = res.PagingToken
```
jyecusch, line 380: If a fetch error occurs here, what would `res.Documents` return? Could it result in a nil pointer dereference and panic?

tjholm: Have put in an error only iterator in the initial state fetch case, to avoid this.

jyecusch (approved):

jyecusch (approved):

jyecusch (approved):

davemooreuws (approved):

PR #175, opened 2021-09-17
Mongodb streaming support

tjholm:


pkg/plugins/document/mongodb/mongodb.go
```diff
@@ -333,62 +330,119 @@ func (s *MongoDocService) Query(collection *document.Collection, expressions []d
...
+}
 
-	return queryResult, nil
+func mongoDocToDocument(coll *document.Collection, cursor *mongo.Cursor) (*document.Document, error) {
+	var docSnap map[string]interface{}
+
+	err := cursor.Decode(&docSnap)
+
+	if err != nil {
+		return nil, fmt.Errorf("error decoding mongo document")
```
medgar-nitric, line 415: I think we should just return the error here, to avoid eating error the root cause.

pkg/plugins/document/mongodb/mongodb.go
```diff
@@ -333,62 +330,119 @@ func (s *MongoDocService) Query(collection *document.Collection, expressions []d
...
+	return func() (*document.Document, error) {
+		if cursorErr != nil {
+			return nil, cursorErr
+		}
 
-			queryResult.PagingToken = map[string]string{
-				"pagingTokens": tokens,
+		if cursor.Next(s.context) {
+			// return the next document
+			return mongoDocToDocument(collection, cursor)
```
medgar-nitric, line 393: Please use error function:

```
doc, err = mongoDocToDocument(collection, cursor)
if err != nil {
	return nil, newErr(
		codes.Internal,
		"error decoding mongo document",
		err,
	)
}

return doc, nil
```

pkg/plugins/document/mongodb/mongodb.go
```diff
@@ -311,17 +287,38 @@ func (s *MongoDocService) Query(collection *document.Collection, expressions []d
...
+
+func (s *MongoDocService) Query(collection *document.Collection, expressions []document.QueryExpression, limit int, pagingToken map[string]string) (*document.QueryResult, error) {
+	newErr := errors.ErrorsWithScope(
+		"MongoDocService.Query",
+		map[string]interface{}{
+			"collection": collection,
+		},
+	)
+
+	if colErr, expErr := document.ValidateQueryCollection(collection), document.ValidateExpressions(expressions); colErr != nil || expErr != nil {
```
medgar-nitric, line 309: smile :)

davemooreuws (approved):

pkg/plugins/document/mongodb/mongodb.go
tjholm: done

pkg/plugins/document/mongodb/mongodb.go
tjholm: done

medgar-nitric (approved):

PR #176, opened 2021-09-20
Dynamo DB streaming.

tjholm:


pkg/plugins/document/dynamodb/dynamodb.go
```diff
@@ -285,69 +322,91 @@ func (s *DynamoDocService) Query(collection *document.Collection, expressions []
...
-		remainingLimit := limit - len(queryResult.Documents)
+	var tmpLimit = limit
+	var documents []document.Document
+	var pagingToken map[string]string
 
-		// If more results available, perform additional queries
-		for remainingLimit > 0 &&
-			(queryResult.PagingToken != nil && len(queryResult.PagingToken) > 0) {
+	// Initial fetch
+	res, fetchErr := s.query(collection, expressions, limit, nil)
```
jyecusch, line 363: `s.query` is an eager query, with the full limit provided I think this will retrieve all the pages at once. The eager portion of `query` should probably be moved into `Query` so the code can be reused like this.

pkg/plugins/document/dynamodb/dynamodb.go
```diff
@@ -261,6 +262,42 @@ func (s *DynamoDocService) Delete(key *document.Key) error {
...
+
+		if res, err := resFunc(collection, expressions, remainingLimit, queryResult.PagingToken); err != nil {
+			return nil, err
+		} else {
+			queryResult.Documents = append(queryResult.Documents, res.Documents...)
+			queryResult.PagingToken = res.PagingToken
+		}
+
+		remainingLimit = limit - len(queryResult.Documents)
+	}
```
jyecusch, line 296: This is the eager code I mention in the next comment

jyecusch (changes requested):

pkg/plugins/document/dynamodb/dynamodb.go
tjholm: done

pkg/plugins/document/dynamodb/dynamodb.go
tjholm: done

jyecusch (approved):

davemooreuws (approved):

davemooreuws:
LGTM @tjholm just some conflicts need to be resolved

PR #182, opened 2021-09-28
Feature/presigned url gcp

HomelessDinosaur:


pkg/plugins/storage/storage/storage.go
```diff
@@ -179,6 +198,85 @@ func (s *StorageStorageService) Delete(bucket string, key string) error {
...
+		)
+	}
+	jsonKey, err := ioutil.ReadFile(serviceAccount)
+	if err != nil {
+		return "", newErr(
+			codes.Internal,
+			"error reading service account config file at given SERVICE_ACCOUNT_LOCATION",
+			err,
+		)
+	}
```
jyecusch, line 246: I think the credentials can be retrieved using the same method we use to construct the client:

`google.FindDefaultCredentials(ctx, storage.ScopeReadWrite)`

Maybe store them privately on the StorageStorageService struct and reuse them here instead of looking in a custom location.

The returned value has a `credentials.JSON` property, which you can probably use with the remaining code.


Interestingly, other Google Cloud SDKs like Node don't ask for the `PrivateKey` or `GoogleAccessID` when signing URLs. Maybe double-check how they're implemented and we can copy that behaviour.

https://cloud.google.com/storage/docs/access-control/signing-urls-with-helpers#code-samples

pkg/plugins/storage/storage/storage.go
```diff
@@ -179,6 +198,85 @@ func (s *StorageStorageService) Delete(bucket string, key string) error {
...
+	var headers []string
+	if operation == plugin.WRITE { // If its a put operation, need to add additional headers
+		headers = append(headers, "Content-Type:application/octet-stream")
+	} else {
+		operation = plugin.READ
+	}
+	opts := &storage.SignedURLOptions{
+		Scheme:         storage.SigningSchemeV4,
+		Method:         []string{"GET", "PUT"}[operation],
```
jyecusch, line 263: I feel like translation between operations `READ` and `WRITE` to `"GET"` or `"PUT"` should be more explicitly defined instead of relying on the array orders matching.

pkg/plugins/storage/plugin.go
```diff
@@ -25,6 +25,9 @@ const (
 
 func (op Operation) String() string {
 	// The order of this array must match the iota order above.
+	if op != WRITE && op != READ {
```
jyecusch, line 28: If an unknown operation is passed here it should probably return "UNKNOWN" or something similar instead of defaulting to read. Otherwise, error messages could be misleading when they claim a read operation failed when really it was some other unsupported operation. (Whatever you do here we should do the same for the existing AWS S3 implementation).

pkg/plugins/storage/storage/storage.go
```diff
@@ -18,22 +18,41 @@ import (
...
+
+type StorageStorageUtils struct{}
+
+func (u *StorageStorageUtils) JWTConfigFromJSON(jsonKey []byte) (*jwt.Config, error) {
+	return google.JWTConfigFromJSON(jsonKey)
+}
+
+func (u *StorageStorageUtils) SignedURL(bucket string, key string, opts *storage.SignedURLOptions) (string, error) {
+	return storage.SignedURL(bucket, key, opts)
+}
```
jyecusch, line 49: These functions should be private

jyecusch (changes requested):
Looking good, just a few changes/comments

pkg/plugins/storage/plugin.go
asalkeld: maybe use: https://pkg.go.dev/golang.org/x/tools/cmd/stringer

pkg/plugins/storage/storage/storage.go
```diff
@@ -18,22 +18,41 @@ import (
...
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
 )
 
+type StorageUtils interface {
+	JWTConfigFromJSON(jsonKey []byte) (*jwt.Config, error)
+	SignedURL(bucket string, key string, opts *storage.SignedURLOptions) (string, error)
+}
+
+type StorageStorageUtils struct{}
```
asalkeld, line 41: this looks a bit odd (I really don't know the code yet, so please excuse me if I am off base), any way you could have:
```go
type storageUtils struct{} // lower case

// also good to have this as it makes it more obvious what struct implements what interface
var _ StorageUtils = &storageUtils{}
```

PR #184, opened 2021-10-05
Support Multiple Query Parameters

tjholm:


medgar-nitric (approved):

medgar-nitric:
recommend adding `-d` option with makefile go get to reduce warning messages in build:

```
git diff makefile
diff --git a/makefile b/makefile
index e69da27..ca47df6 100644
--- a/makefile
+++ b/makefile
@@ -16,7 +16,7 @@ install:

 install-tools: install
        @echo Installing tools from tools.go
-       @cat ./tools/tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go get %
+       @cat ./tools/tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go get -d %

 clean:
        @rm -rf ./bin/
```

jyecusch (approved):
LGTM, interesting that the raw query string was required for lambda.

PR #189, opened 2021-10-09
fix:faasworker setting response content-type

medgar-nitric:


tjholm:
Definitely in Faas mode (the stack templates don't support anything else).

What version of the membrane was your initial test on? the CLI i've tested with is using v0.12.0-rc.12

Here is a test from v0.12.0-rc.17 (current develop), using a modified version of the go stacks create method:
image

I suspect in the CLI test that you've run above that the membrane version is likely v0.10.2 (latest published version), which was prior to the merge of the multiple HTTP header support, which could point to a backwards compatibility problem with that version of the membrane and the SDKS.

Let me know if the version you're testing with was locked during your CLI test above. It would also be interesting if you observe the same behaviour in a local test on current develop.

PR #193, opened 2021-12-05
Update import path to github.com/nitrictech/nitric

asalkeld:
This is a simple replace of the module path from nitric-dev/membrane to nitrictech/nitric

More can be done as a follow up to simplify the tests.


tjholm (approved):
LGTM

tjholm (changes requested):
Need to do a find and replace on files for integration tests (located under `tests/`).

tjholm (approved):

davemooreuws (approved):

PR #204, opened 2022-01-07
Add runtime handling for new rpc resource and FaaS contracts.

tjholm:


.gitignore
```diff
@@ -26,4 +26,7 @@ interfaces/
 
 mocks/
 
+# Ignore generated mocks
+*.mock.go
+
```
asalkeld, line 31: I think we should check in the mocks

pkg/plugins/gateway/appservice/http.go
```diff
@@ -44,8 +44,8 @@ func handleSubscriptionValidation(ctx *fasthttp.RequestCtx, events []eventgrid.E
 	ctx.Success("application/json", responseBody)
 }
 
-func handleNotifications(ctx *fasthttp.RequestCtx, events []eventgrid.Event, wrkr worker.Worker) {
-	// FIXME: As we are batch handling events in azure
+func handleNotifications(ctx *fasthttp.RequestCtx, events []eventgrid.Event, pool worker.WorkerPool) {
+	// TODO: As we are batch handling events in azure
```
asalkeld, line 48:
```suggestion
	// TODO: As we are batch handling events
```

pkg/worker/grpc_worker.go
```diff
@@ -0,0 +1,285 @@
...
+	// wait for the response
+	// FIXME: Need to handle timeouts here...
+	response := <-returnChan
+
+	topic := response.GetTopic()
+
+	if topic == nil {
+		// Fatal error in this case
+		// We don't have the correct response type for this handler
+		return fmt.Errorf("Fatal: Error handling event, incorrect response received from function")
```
asalkeld, line 269: should probably create a custom FatalError type ..

pkg/worker/route_worker.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/nitrictech/nitric/pkg/utils"
+
+	"github.com/nitrictech/nitric/pkg/triggers"
+
+	pb "github.com/nitrictech/nitric/interfaces/nitric/v1"
```
asalkeld, line 25: clean up imports (no need for spacing)

.gitignore
tjholm: I'd like to understand the reasoning behind this more. If code generation is idempotent for a given set of sources/interfaces what would the value of checking this code in be?

I can understand where we might need this to make the repo `go get`able (in the case of the go SDK), but currently this repo serves binary only distribution and not really intended to be used as a lib at this point in time.

If there are other genuine pros as well I'd be keen to chat about them, but I could see versioing this code introducing some pain points as well.

asalkeld: Every golang project I have been involved with has checked in anything generated.

https://www.reddit.com/r/golang/comments/8mbi47/do_you_commit_all_the_generated_code_in_your/
https://stackoverflow.com/questions/56415527/should-a-developer-commit-go-generated-code

Basically, I think it's a good idea to do what most go developers are used to...

tjholm: I'll raise a new ticket to look at this so we can take care of the CI/CD at the same time as well as versioning the generated sources.

pkg/worker/route_worker.go
tjholm: done.

pkg/plugins/gateway/appservice/http.go
tjholm: done.

pkg/worker/grpc_worker.go
tjholm: Have put in a new errors package and add in a new FatalError type for now (can add additional types as we refactor).

asalkeld (approved):

davemooreuws (approved):

PR #205, opened 2022-01-12
Route to schedules based on topic naming convention

tjholm:


davemooreuws (approved):

pkg/worker/schedule_worker.go
```diff
@@ -33,9 +34,9 @@ func (s *ScheduleWorker) HandlesHttpRequest(trigger *triggers.HttpRequest) bool
 }
 
 func (s *ScheduleWorker) HandlesEvent(trigger *triggers.Event) bool {
-	// TODO: Determine if event should be handled by using convention for
-	// this schedule
-	return false
+	topicName := strings.ToLower(strings.ReplaceAll(trigger.Topic, " ", "-"))
```
jyecusch, line 37: Could we pull this name conversion into a function so it's self-documenting and reusable? Otherwise, just a comment with the intention would be useful, otherwise the reason for this string conversion could be forgotten and unclear

jyecusch (changes requested):

pkg/worker/schedule_worker.go
tjholm: Once https://github.com/nitrictech/nitric/pull/206 is in we can create a common method that could actually be shared between the membrane and CLI as we need to use this convention for deployment as well atm.

asalkeld (approved):

jyecusch (approved):

PR #208, opened 2022-01-14
Preserve API name as part of route worker and make it readable.

tjholm:
Preserving API name metadata for route workers. This will allow us to do named API sub routing in single membrane gateway scenarios e.g. nitric run.


davemooreuws (approved):

asalkeld (approved):
lgtm

PR #211, opened 2022-01-17
feat: allow getting all matching workers from pool

jyecusch:
This lets us get all workers when running locally and retrieving all subscribers for an event.

### DIFF
--- a/pkg/worker/pool.go
+++ b/pkg/worker/pool.go
@@ -170,10 +170,6 @@ func (p *ProcessPool) GetWorkers(opts *GetWorkerOptions) []Worker {
 	if opts.Http != nil {
 		ws := p.getHttpWorkers()
 
-		if opts.Filter != nil {
-			ws = filterWorkers(ws, opts.Filter)
-		}
-
 		for _, w := range ws {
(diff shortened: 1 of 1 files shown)


pkg/worker/pool.go
```diff
@@ -158,6 +159,45 @@ func filterWorkers(ws []Worker, f func(w Worker) bool) []Worker {
...
+	defer p.workerLock.Unlock()
+
+	workers := make([]Worker, 0)
+
+	if opts.Http != nil {
+		ws := p.getHttpWorkers()
+
+		if opts.Filter != nil {
+			ws = filterWorkers(ws, opts.Filter)
+		}
```
tjholm, line 175: Because we don't have a case where we need to call the actual hander for a single worker we could move this to the end of the function I think

tjholm (changes requested):
LGTM, just one comment on the filter placement.

pkg/worker/pool.go
jyecusch: done

davemooreuws (approved):

tjholm (approved):

PR #212, opened 2022-01-19
feat: update event message to match CloudEvents spec

jyecusch:
Not fully compatible with cloud events yet, but closer. This shouldn't be a breaking change.

### DIFF
--- a/contracts/proto/event/v1/event.proto
+++ b/contracts/proto/event/v1/event.proto
@@ -63,7 +63,11 @@ message NitricEvent {
   // A Unique ID for the Nitric Event
   string id = 1;
   // A content hint for the events payload
-  string payload_type = 2;
+  string type = 2;
   // The payload of the event
-  google.protobuf.Struct payload = 3;
+  oneof data {
+    bytes binary_data = 4;
+    string text_data = 5;
+    google.protobuf.Struct proto_data = 3;
+  }
 }


contracts/proto/event/v1/event.proto
```diff
@@ -63,7 +63,11 @@ message NitricEvent {
...
   // A content hint for the events payload
-  string payload_type = 2;
+  string type = 2;
   // The payload of the event
-  google.protobuf.Struct payload = 3;
+  oneof data {
+    bytes binary_data = 4;
+    string text_data = 5;
+    google.protobuf.Struct proto_data = 3;
+  }
```
asalkeld, line 72: if you wanted to be backwards compatible you could just have all fields, then just use the populated data

contracts/proto/event/v1/event.proto
```diff
@@ -63,7 +63,11 @@ message NitricEvent {
...
   string id = 1;
   // A content hint for the events payload
-  string payload_type = 2;
+  string type = 2;
   // The payload of the event
-  google.protobuf.Struct payload = 3;
+  oneof data {
+    bytes binary_data = 4;
+    string text_data = 5;
+    google.protobuf.Struct proto_data = 3;
```
tjholm, line 71: maybe change this to struct_data or object_data?

jyecusch: I took the name from the CloudEvents proto spec for consistency https://github.com/cloudevents/spec/blob/v1.0.1/protobuf-format.md#3-data maybe we should just use this

tjholm (approved):

tjholm (changes requested):
need to generate new server/client code and update handlers.

PR #213, opened 2022-02-21
chore: Use log.Default().Print* functions instead of fmt.Print*

asalkeld:
This will make it easier for the cli to capture the output.

### DIFF
--- a/pkg/membrane/membrane.go
+++ b/pkg/membrane/membrane.go
@@ -15,7 +15,9 @@
 package membrane
 
 import (
+	"errors"
 	"fmt"
+	"log"
 	"net"
 	"os"
 	"os/exec"
@@ -103,9 +105,9 @@ type Membrane struct {
 	pool worker.WorkerPool
 }
 
-func (s *Membrane) log(log string) {
+func (s *Membrane) log(msg string) {
 	if !s.suppressLogs {
-		fmt.Println(log)
+		log.Default().Println(msg)
 	}
 }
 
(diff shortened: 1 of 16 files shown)


pkg/membrane/membrane.go
```diff
@@ -141,15 +143,15 @@ func (s *Membrane) startChildProcess() error {
...
+	log.Default().Println("Starting Child Process")
 	childProcess := exec.Command(s.childCommand[0], s.childCommand[1:]...)
 	childProcess.Stdout = os.Stdout
 	childProcess.Stderr = os.Stderr
 	applicationError := childProcess.Start()
 
 	// Actual panic here, we don't want to start if our userland code cannot successfully start
 	if applicationError != nil {
-		return fmt.Errorf("There was an error starting the child process: %v", applicationError)
+		return fmt.Errorf("there was an error starting the child process: %w", applicationError)
```
HomelessDinosaur, line 154: Could this be a log.Default().Fatal if its an actual panic?

pkg/membrane/membrane.go
```diff
@@ -191,7 +193,7 @@ func (s *Membrane) Start() error {
 	}
 	lis, err := net.Listen("tcp", s.serviceAddress)
 	if err != nil {
-		return fmt.Errorf("Could not listen on configured service address: %v", err)
+		return fmt.Errorf("could not listen on configured service address: %w", err)
```
HomelessDinosaur, line 196: Could this be log.Default().Printf?

asalkeld: it's returning an error. that should bubble up to a log at some point.

pkg/membrane/membrane.go
asalkeld: no, it will return  an error to the caller. We shouldn't assume failure handling

HomelessDinosaur (approved):

tjholm (approved):

PR #236, opened 2022-03-04
AWS Resource Discovery

tjholm:
Implement more robust discovery of AWS resources.

Have tested this with the chess-api, would be good to do some more tests.


jyecusch (approved):
LGTM, gave it a test with topics and buckets

asalkeld (approved):
lgtm, lots of changes..

PR #240, opened 2022-03-06
Add security definitions to APIs and API Workers.

tjholm:
Proposal for API/API worker security definitions.

Security definitions are defined at the API level, with routes and workers opting into named security definitions, from their parent API.

potential example:
```typescript
import { api } from "@nitric/sdk";

// example auth0 issuer configuration
const public = api('public', {
  security: {
     'client': api.JWTAuthorizer({
        issuer: 'https://YOUR-ACCOUNT-NAME.auth0.com',
        audiences: ['test']
     })
  }
});

// Not sure on the best way we could reflect this for the API routes
// example here looks pretty messy
api.get('/test', async ctx => {
  
}, {
  // map to the above authorizer by name
  security: 'client'
});
```


asalkeld (commented):
Seems ok, I assume the token will be placed in the context so the user can then extract out their
claim info?

tjholm:
> Seems ok, I assume the token will be placed in the context so the user can then extract out their claim info?
:+1: 

All the original headers from HTTP requests will still be available/passed through

davemooreuws (approved):

tjholm:
Need to consider local run as well for these changes.

jyecusch:
Could we make the reference to the authorizer used in the routes a variable, instead of a string so things like refactoring and auto-complete are easier? Particularly, when the APIs get complex and defined across multiple source files.

e.g. 

```typescript
import { api } from "@nitric/sdk";

const public = api('public', {
  security: {
     'client': api.JWTAuthorizer({
        issuer: 'https://YOUR-ACCOUNT-NAME.auth0.com',
        audiences: ['test']
     })
  }
});

public.get('/test', async ctx => {
  
}, {
  // map to the above authorizer
  security: public.security.client
});
```

Also, should we consider config that enforces the security for all routes - set and forget at the API level?

tjholm:
> Could we make the reference to the authorizer used in the routes a variable, instead of a string so things like refactoring and auto-complete are easier? Particularly, when the APIs get complex and defined across multiple source files.
> 
> e.g.
> 
> ```ts
> import { api } from "@nitric/sdk";
> 
> const public = api('public', {
>   security: {
>      'client': api.JWTAuthorizer({
>         issuer: 'https://YOUR-ACCOUNT-NAME.auth0.com',
>         audiences: ['test']
>      })
>   }
> });
> 
> public.get('/test', async ctx => {
>   
> }, {
>   // map to the above authorizer
>   security: public.security.client
> });
> ```
> 
> Also, should we consider config that enforces the security for all routes - set and forget at the API level?

We should definitely do this! We can probably use the generics on the parent resource to imply the keys supplied in the security definitions to ensure they are properly typed (e.g. `KeyOf<DefObject>`).

tjholm:
new javascript example (including scopes):
```typescript
import { api, jwt } from '../src';

const testApi = api('test', {
    securityDefinitions: {
        user: jwt({
            issuer: "https://dev-fn1x0c3o.us.auth0.com/",
            audiences: ['testing'],
        }),
    },
    security: {
        "user": []
    },
});

testApi.get("/hello/:name", async (ctx) => {
    ctx.res.body = `Hello`
});

testApi.get("/test/:name", async (ctx) => {
    ctx.res.body = `No Security`
}, { security: {} });
```

jyecusch (approved):

PR #241, opened 2022-03-06
Add function level configuration

tjholm:
Proposal to add function level configuration to config as code.

Currently allows for:
 - Naming
 - Set desired Memory (CPUs will be driven by the selected memory depending on provider availability.


contracts/proto/resource/v1/resource.proto
```diff
@@ -63,6 +64,20 @@ message TopicResource {}
...
+  // name of the function resource (defaults based on file name)
+  string name = 1;
+  // The desired cpu count of the function
+  // The default will be the minimum available on the provider being deployed to
+  // and the minimum required to support requested memory
+  int32 cpus = 2;
+  // The desired memory of the function in mebibytes
+  // the default will be the minimum available on the provider
+  // being deployed to
+  int32 memory = 3;
```
asalkeld, line 78: buildScripts ? @davemooreuws was asking for them..
min/max scale limits?

tjholm: @asalkeld both good suggestions, I think we should raise these as separate PRs following this one.

jyecusch: Yeah, let's have that chat. Adding build scripts could help with the limitations with Prisma use right now too - i.e. we could run `npx prisma generate`

asalkeld (approved):

tjholm:
thinking more about this one, was wondering if we need to spend some more time fleshing out the infrastructure/code infrastructure split, would memory configuration be better defined in a configuration file?

jyecusch:
> thinking more about this one, was wondering if we need to spend some more time fleshing out the infrastructure/code infrastructure split, would memory configuration be better defined in a configuration file?

I think it would be better to live in the stack files, so you can set more memory, etc. in prod v. non-prod environments. One challenge we have with that is the separation of code/config again. For example, if the code defining a function is changed/deleted and the config is still present errors could be introduced, can we prevent that?

tjholm:
closing for now, will revisit if we need this at code level, but will look at defining it as part of a runtime as an alternative.

PR #253, opened 2022-03-21
fix: converting slice or maps of specific types (non interface)

asalkeld:
part of nitrictech/go-sdk#44

Once this is merged, I'll use this function utlis.NewStruct() in go-sdk.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,6 +33,7 @@ require (
 	github.com/googleapis/gax-go/v2 v2.1.1
 	github.com/lyft/protoc-gen-star v0.6.0 // indirect
 	github.com/mitchellh/mapstructure v1.4.2
+	github.com/nitrictech/protoutils v0.0.0-20220321024151-14f05ec4cd27
 	github.com/onsi/ginkgo v1.16.5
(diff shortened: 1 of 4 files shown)


tjholm (changes requested):
@asalkeld looks good!

My one concern is using this code as a direct dependency in the go-sdk, as they both contain potentially conflicting versions of our generated SDK code (the client SDK has some minor differences that make it easier to mock, but less safe to use for writing a server).

Would it be worth creating a separate lib to manage this? I can see us creating additional proto related tools and utilities in future.

asalkeld:
> @asalkeld looks good!
> 
> My one concern is using this code as a direct dependency in the go-sdk, as they both contain potentially conflicting versions of our generated SDK code (the client SDK has some minor differences that make it easier to mock, but less safe to use for writing a server).
> 
> Would it be worth creating a separate lib to manage this? I can see us creating additional proto related tools and utilities in future.

Here it is: github.com/nitrictech/protoutils

tjholm (approved):

jyecusch (approved):
👍

PR #259, opened 2022-03-28
build: install protoc with a fixed version so the generated files don't keep changing

asalkeld:


tools/tools.mk
```diff
@@ -0,0 +1,66 @@
...
+endif
+
+ifndef PROTOC_PLATFORM
+    $(error unsupported platform $(UNAME_S):$(UNAME_P))
+endif
+
+TOOLS_DIR := ./tools
+TOOLS_BIN := $(TOOLS_DIR)/bin
+
+PROTOC_VERSION := 3.19.1
```
asalkeld, line 38: this is the main thing here ^. We will all have the same version..

jyecusch (approved):

tjholm (approved):

Issue #256, opened 2022-03-22
Increase the coverage of testing against the Nitric Membrane from 50% to 70%


raksiv:

raksiv:
The goal here is to improve our test coverage, not to hit a number specifically.

raksiv:
Closing - completed by Angus.

PR #261, opened 2022-03-29
More unit test and add more linters

asalkeld:
part of https://github.com/nitrictech/nitric/issues/256


pkg/providers/aws/plugin.go
```diff
@@ -25,26 +25,13 @@ import (
...
-
-	if err != nil {
-		return nil, err
-	}
-
-	return &AWSServiceFactory{
-		provider: provider,
-	}, nil
-}
-
```
tjholm, line 47: If the code in these is dead, we can probably remove these.

Initially wanted to look at using something like `wire` to do compile time dependency injection for building the different provider membranes, but we can clean out this dead code for now.

pkg/adapters/grpc/faas_grpc.go
```diff
@@ -88,8 +88,10 @@ func (s *FaasServer) TriggerStream(stream pb.FaasService_TriggerStreamServer) er
 	log.Default().Println("FaaS stream closed, removing worker")
 
 	// Worker is done so we can remove it from the pool
-	s.pool.RemoveWorker(wrkr)
-
+	rwErr := s.pool.RemoveWorker(wrkr)
+	if err == nil {
+		err = rwErr
+	}
 	return err
```
tjholm, line 95: good that error handling is being added here, but the logic seems a little confusing, if we experience an error calling pool.RemoveWorker, but we had another error before, we're justing throwing away the new error?

asalkeld: yeah, I wasn't sure what the priority was.. of if they were dependent. I could have an error list here..

davemooreuws (approved):
Nice work Angus! Approved, Tims suggestions can go in here or new pr.

pkg/adapters/grpc/faas_grpc.go
asalkeld: improvement comming

pkg/plugins/storage/boltdb/storage.go
```diff
@@ -166,21 +137,6 @@ func (s *BoltStorageService) Delete(bucket string, key string) error {
...
-		)
-	}
-	if key == "" {
-		return newErr(
-			codes.InvalidArgument,
-			"provide non-blank key",
-			nil,
-		)
-	}
-
```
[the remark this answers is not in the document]
asalkeld: removed

tjholm (approved):

PR #263, opened 2022-04-04
fix builds on arch.

tjholm:
Updating uname call for getting architecture for protoc distribution.

### DIFF
--- a/tools/tools.mk
+++ b/tools/tools.mk
@@ -12,7 +12,7 @@ ifeq ($(OS),Windows_NT)
 else
     UNAME_S := $(shell uname -s)
     ifeq ($(UNAME_S),Linux)
-        UNAME_P := $(shell uname -p)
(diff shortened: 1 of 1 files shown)


tjholm:
@asalkeld & @raksiv will need to make sure this change still works on your machines as well.

jyecusch (approved):

tools/tools.mk
```diff
@@ -12,7 +12,7 @@ ifeq ($(OS),Windows_NT)
 else
     UNAME_S := $(shell uname -s)
     ifeq ($(UNAME_S),Linux)
-        UNAME_P := $(shell uname -p)
+        UNAME_P := $(shell uname -m)
```
asalkeld, line 15:
```bash
/> uname -m
x86_64

```

asalkeld (approved):

PR #274, opened 2022-05-02
Add cloud event worker type

tjholm:
Adds a worker type that responds to cloud events.


tjholm:
TODO: Need to discuss format for sources for standard nitric event registration for infrastructure provisioning (or decide if we separate out the binding definition.

A good example of this would be replacing our current topic subscription infrastructure with a cloud event binding. Where we define the source of the event to be the nitric topic:

example:
`nitric:topic:my-topic`

tjholm:
closing as stale

PR #277, opened 2022-05-10
feat(cloudrun): give a better error if we can't find the topic name

asalkeld:
Also many messages have an "id", so use the payloadType as well to determine
if the message is a nitric one.

### DIFF
--- a/pkg/plugins/gateway/cloudrun/http.go
+++ b/pkg/plugins/gateway/cloudrun/http.go
@@ -47,24 +47,36 @@ func middleware(ctx *fasthttp.RequestCtx, pool worker.WorkerPool) bool {
 	if err := json.Unmarshal(bodyBytes, &pubsubEvent); err == nil && pubsubEvent.Subscription != "" {
 		// We have an event from pubsub here...
 
+		topicName, ok := pubsubEvent.Message.Attributes["x-nitric-topic"]
+		if !ok {
+			ctx.Error("Could not find topicName for event", 400)
+			return false
+		}
+
 		// need to determine if the underlying data is a nitric event
 		var event *triggers.Event
 		messageJson := &ep.NitricEvent{}
+
 		// Check if it's a nitric event
-		if err := json.Unmarshal(pubsubEvent.Message.Data, messageJson); err == nil && messageJson.ID != "" {
+		err := json.Unmarshal(pubsubEvent.Message.Data, messageJson)
+		if err == nil && messageJson.ID != "" && messageJson.PayloadType != "" {
 			// reserialize the nitric event payload
-			payload, _ := json.Marshal(messageJson.Payload)
+			payload, err := json.Marshal(messageJson.Payload)
+			if err != nil {
+				ctx.Error("Could not marshal event payload: "+err.Error(), 400)
+				return false
+			}
 
 			event = &triggers.Event{
 				ID:      messageJson.ID,
-				Topic:   pubsubEvent.Message.Attributes["x-nitric-topic"],
+				Topic:   topicName,
 				Payload: payload,
 			}
 		} else {
 			event = &triggers.Event{
 				ID: pubsubEvent.Message.ID,
 				// Set the topic
-				Topic: pubsubEvent.Message.Attributes["x-nitric-topic"],
+				Topic: topicName,
 				// Set the original full payload payload
 				Payload: pubsubEvent.Message.Data,
 			}
@@ -73,7 +85,6 @@ func middleware(ctx *fasthttp.RequestCtx, pool worker.WorkerPool) bool {
 		wrkr, err := pool.GetWorker(&worker.GetWorkerOptions{
 			Event: event,
 		})
-
 		if err != nil {
 			ctx.Error("Could not find handle for event", 500)
 			return false


pkg/plugins/gateway/cloudrun/http.go
```diff
@@ -47,24 +47,37 @@ func middleware(ctx *fasthttp.RequestCtx, pool worker.WorkerPool) bool {
 	if err := json.Unmarshal(bodyBytes, &pubsubEvent); err == nil && pubsubEvent.Subscription != "" {
 		// We have an event from pubsub here...
 
+		topicName, ok := pubsubEvent.Message.Attributes["x-nitric-topic"]
+		if !ok {
+			fmt.Println("Could not find topicName for event")
+			ctx.Error("Could not find topicName for event", 400)
+			return true
```
HomelessDinosaur, line 54: Looks great, just one question. Should this return false if you don't want the event to continue processing?

pkg/plugins/gateway/cloudrun/http.go
```diff
@@ -47,24 +47,37 @@ func middleware(ctx *fasthttp.RequestCtx, pool worker.WorkerPool) bool {
...
 		// Check if it's a nitric event
-		if err := json.Unmarshal(pubsubEvent.Message.Data, messageJson); err == nil && messageJson.ID != "" {
+		err := json.Unmarshal(pubsubEvent.Message.Data, messageJson)
+		if err == nil && messageJson.ID != "" && messageJson.PayloadType != "" {
 			// reserialize the nitric event payload
-			payload, _ := json.Marshal(messageJson.Payload)
+			payload, err := json.Marshal(messageJson.Payload)
+			if err != nil {
+				ctx.Error("Could not marshal event payload: "+err.Error(), 400)
+				return true
```
HomelessDinosaur, line 68: As above.

HomelessDinosaur (changes requested):

pkg/plugins/gateway/cloudrun/http.go
asalkeld: @tjholm can you chime in here? I thought you suggested this?

tjholm: correct, if we've already delegated a HTTP response (like using `ctx.Error`) then we shouldn't have the base plugin continue processing. If you want the base plugin to attempt to continue searching for a worker to handle the request you can directly return true, but we shouldn't also return a response).

HomelessDinosaur (approved):

pkg/plugins/gateway/cloudrun/http.go
```diff
@@ -47,24 +47,36 @@ func middleware(ctx *fasthttp.RequestCtx, pool worker.WorkerPool) bool {
...
+		}
+
 		// need to determine if the underlying data is a nitric event
 		var event *triggers.Event
 		messageJson := &ep.NitricEvent{}
+
 		// Check if it's a nitric event
-		if err := json.Unmarshal(pubsubEvent.Message.Data, messageJson); err == nil && messageJson.ID != "" {
+		err := json.Unmarshal(pubsubEvent.Message.Data, messageJson)
+		if err == nil && messageJson.ID != "" && messageJson.PayloadType != "" {
```
tjholm, line 62: payload type in some cases may be optional (we don't actually enforce its delivery)

pkg/plugins/gateway/cloudrun/http.go
```diff
@@ -47,24 +47,36 @@ func middleware(ctx *fasthttp.RequestCtx, pool worker.WorkerPool) bool {
 	if err := json.Unmarshal(bodyBytes, &pubsubEvent); err == nil && pubsubEvent.Subscription != "" {
 		// We have an event from pubsub here...
 
+		topicName, ok := pubsubEvent.Message.Attributes["x-nitric-topic"]
+		if !ok {
+			ctx.Error("Could not find topicName for event", 400)
+			return false
+		}
+
```
tjholm, line 55: Wasn't part of this change to retrieve the topic name from a query parameter if it's available

tjholm (changes requested):

pkg/plugins/gateway/cloudrun/http.go
asalkeld: it was the idea, but it does not work as it changes the signature and breaks the authorizer on the function.
So i at least wanted a better error.

pkg/plugins/gateway/cloudrun/http.go
asalkeld: ok, what else do you suggest. because almost all objects have an "id". so that is really not a good indication of it being a nitric event.

pkg/plugins/gateway/cloudrun/http.go
tjholm: Is there an alternative we could use that would not impact the authorizer? (e.g. using path instead?).

Another alternative could be to look at binding using eventarcs instead as well.

pkg/plugins/gateway/cloudrun/http.go
tjholm: At the moment I wouldn't apply any changes. Currently looking into using the cloud events spec for our eventing and would end up deprecating the "Nitric Event" anyway.

PR #292, opened 2022-06-16
feat: increase read buffer size to accomodate large bearer token headers

HomelessDinosaur:


tjholm:
@HomelessDinosaur can you check to see if the mod/sum is up to date for this PR. Looks like there is an issue with dependencies in testing.

raksiv (approved):

Issue #287, opened 2022-06-02
Design subscription DeadLetter queues support


raksiv:
It should be possible to configure dead-letter queues for topic subscriptions so that when an event cannot be delivered to the intended subscriber, it can end up in a queue for processing.

The dead letter queue is a service implementation to store messages that meet one or more of the following criteria:

Dead letter queue storing of these messages allows developers to look for common patterns and potential software problems.

- Design the interface for the DL queue in the Nitric Server (Sprint 1)

jyecusch:
The initial design was to use a `queue` resource as the dead-letter destination. However, Azure Event Grid doesn't support dead-letter messages being sent to queues, only to storage as blobs (see: https://docs.microsoft.com/en-us/azure/event-grid/manage-event-delivery). This leaves us with a few options:

- Don't support dead-letter queues for Azure
- Process the dead-letter messages from storage and send them to a queue (i.e. we provide a service to do this)
- Provide inconsistent behaviour between clouds for dead-letters (e.g. set the name of the dead-letter target, which will be a queue in AWS, topic in PubSub, storage on Azure)
- Make dead-letter targets a unique resource. Since we _could_ subscribe a function to the write events of the dead-letter blob container in Azure. We could make it possible to do something like `deadletter.subscribe(...)` and provide similar behaviour across the clouds, even though different underlying resources would be used.

jyecusch:
@tjholm suggests option 4 is probably the best. We could also make the dead-letter nitric resource queryable - so instead of subscribing you could run a schedule (e.g. daily) to pull the latest events off the dead-letter queue.

raksiv:
That seems like the most flexible approach in the options.

PR #295, opened 2022-06-27
Feat/deadletter for subscriptions

jyecusch:
Closes: https://github.com/nitrictech/nitric/issues/287


contracts/proto/faas/v1/faas.proto
```diff
@@ -71,6 +71,8 @@ message ApiWorker {
 
 message SubscriptionWorker {
   string topic = 1;
+  // optional dead-letter queue where events should be sent if they can't be delivered to/processed by this worker.
+  string dead_letter_queue = 2;
```
tjholm, line 75: changing `dead_letter_queue` -> `dead_letter` here might be better wording for supporting different sink types for deadletter messages.

pkg/api/nitric/v1/document.pb.validate.go
```diff
@@ -32,7 +31,6 @@ var (
 	_ = (*url.URL)(nil)
 	_ = (*mail.Address)(nil)
 	_ = anypb.Any{}
-	_ = sort.Sort
```
tjholm, line 35: @jyecusch can you double check the version of protoc/protoc-gen-go you're using for this? It looks like there is a mismatch with the version we're using for CI/CD. I think @asalkeld implmeneted some install scripts in our makefile for this to ensure the versions are consistent, so it looks like we'll need to double check this.

contracts/proto/faas/v1/faas.proto
jyecusch: done

pkg/api/nitric/v1/document.pb.validate.go
jyecusch: I have 3.19.4 installed, it looks like the install script is trying to install 3.19.1

davemooreuws (approved):
LGTM

asalkeld (commented):
seems ok to me, but not exactly sure what the intended behavior is

PR #320, opened 2022-09-08
Add healthcheck to lambda gateway.

tjholm:
Adds a healthcheck we can invoke during deployments to ensure that the lambda is going to work.


asalkeld:
lgtm, just test not passing

asalkeld (approved):

HomelessDinosaur (approved):

PR #330, opened 2022-10-05
Resource details API

tjholm:
WIP currently, posting for discussion and additional proposals


contracts/proto/resource/v1/resource.proto
```diff
@@ -118,4 +121,25 @@ enum Action {
...
+message ResourceDetailsRequest {
+  Resource resource = 1;
+}
+
+message ResourceDetailsResponse {
+  // The identifier of the resource
+  string id = 1;
+  // A urn for the resources provider broken up as <provider>:<resource_type>
+  // e.g. aws:ApiGateway or gcp:ApiGateway or azure:ApiManagement
+  string provder = 2;
```
asalkeld, line 139:
```suggestion
  string provider = 2;
```

jyecusch:
Looks good, do you have a proposal for how the SDK interaction might look?

tjholm:
> Looks good, do you have a proposal for how the SDK interaction might look?

```javascript
```

raksiv:
> > Looks good, do you have a proposal for how the SDK interaction might look?
> 
> Good question, we could tie in an abstract method with generic types (where available) into all base resources?
> 
> ```js
> import { api } from '@nitric/sdk';
> 
> const publicApi = api('public');
> 
> publicApi.get('/hello', () => {
>    const { url } = await publicApi.details();
>    
>    // we could also just add shortcuts for the above as well
>    // or as a synthetic accessor that returns a promise
>    const otherUrl = await publicApi.url();
>    
>    return `Hello from ${url}`;
> });
> ```

The SDK proposal looks good to me - 

Making it as easy as possible to access things like 'url, name, identifier' which are the most common requests seems like the best DX, anything else can come from details().

HomelessDinosaur (approved):

PR #336, opened 2022-10-18
GCP resources API

tjholm:
Add Details API for GCP resources.

Currently only implements APIs as requested.


pkg/providers/gcp/core/provider.go
```diff
@@ -53,6 +64,49 @@ func createMetadataRequest(uri string) (*http.Request, error) {
...
+			Detail: common.ApiDetails{
+				URL: fmt.Sprintf("https://%s", gw.DefaultHostname),
+			},
+		}, nil
+	} else {
+		return nil, err
+	}
+}
+
+func (g *gcpProviderImpl) Details(typ common.ResourceType, name string) (*common.DetailsResponse[any], error) {
```
HomelessDinosaur, line 101:
```suggestion
func (g *gcpProviderImpl) Details(type common.ResourceType, name string) (*common.DetailsResponse[any], error) {
```

pkg/providers/gcp/core/provider.go
```diff
@@ -53,6 +64,49 @@ func createMetadataRequest(uri string) (*http.Request, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	gws := g.apiClient.ListGateways(context.TODO(), &apigatewaypb.ListGatewaysRequest{
+		Parent: fmt.Sprintf("projects/%s/locations/%s", projectName, g.region),
+		Filter: filter(g.stackName, name),
+	})
+
+	// there should only be a single entry in this array, we'll grab the first and then brake
```
HomelessDinosaur, line 86:
```suggestion
	// there should only be a single entry in this array, we'll grab the first and then break
```

HomelessDinosaur (changes requested):

pkg/providers/gcp/core/provider.go
tjholm: `type` is a reserved keyword in golang

davemooreuws (approved):
LGTM

HomelessDinosaur (approved):

Issue #329, opened 2022-10-04
Distributed tracing


jyecusch:
### Suggestion

Investigate the best approach to provide automatic inclusion of distributed tracing in the framework. What tools and standards already exist that we should use/integrate with/support?

### Value

Tracing in distributed systems is useful to fully understand the impact of a request and identify issues, bottlenecks or perform debugging.

### Alternatives

The current option is to implement your own solution, but we've had several requests to provide out of the box support for this since it's such a common need.

tjholm:
https://opentelemetry.io/

PR #342, opened 2022-10-25
Implement telemetry

asalkeld:
part of https://github.com/nitrictech/nitric/issues/329


pkg/plugins/gateway/base_http/http.go
```diff
@@ -40,8 +43,20 @@ type BaseHttpGateway struct {
 	mw HttpMiddleware
 }
 
-func (s *BaseHttpGateway) httpHandler(pool worker.WorkerPool) func(ctx *fasthttp.RequestCtx) {
+func (s *BaseHttpGateway) httpHandler(unWrappedPool worker.WorkerPool) func(ctx *fasthttp.RequestCtx) {
 	return func(ctx *fasthttp.RequestCtx) {
+		pool := &worker.InstrumentedWorkerPool{
+			WorkerPool: unWrappedPool,
+			Wrapper:    worker.InstrumentedWorkerFn(span.FromFastHttp(ctx), true, false),
+		}
```
tjholm, line 51: should we be re-wrapping this on every request? We can update the worker methods to accept context to potentially avoid this.

pkg/plugins/gateway/lambda/lambda.go
```diff
@@ -253,15 +255,30 @@ func (s *LambdaGateway) handle(ctx context.Context, data map[string]interface{})
...
 	// s.finished = make(chan int)
-	s.pool = pool
+
 	// Here we want to begin polling lambda for incoming requests...
-	s.runtime(s.handle)
+	s.runtime(func(ctx context.Context, data map[string]interface{}) (interface{}, error) {
+		s.pool = &worker.InstrumentedWorkerPool{
+			WorkerPool: pool,
+			Wrapper:    worker.InstrumentedWorkerFn(spanFromContext(ctx), false, true),
+		}
```
tjholm, line 271: should we be re-wrapping this on every request? We could update the handler signatures on the base worker pools to accepts context, and add the span creating in the instrumented worker pools handle methods, so we can wrap once for all contexts.

pkg/pm/processmanager.go
```diff
@@ -0,0 +1,142 @@
...
+		log.Default().Printf("Starting: %s", pm.preProcesses[i].Command[0])
+
+		err := pm.preProcesses[i].cmd.Start()
+		if err != nil {
+			return errors.WithMessagef(err, "there was an error starting the process %s", pm.preProcesses[i].Command[0])
+		}
+	}
+
+	return nil
+}
```
tjholm, line 90: Is it necessary to differentiate between pre-processing and the user space process? There is a bit here we can de-duplicate if we know that the consuming process can just give the order to the process manager.

i.e. processes = [pre..., userspace]

pkg/worker/instrumented_pool.go
```diff
@@ -0,0 +1,33 @@
...
+var _ WorkerPool = &InstrumentedWorkerPool{}
+
+func (iwp *InstrumentedWorkerPool) GetWorker(opts *GetWorkerOptions) (Worker, error) {
+	w, err := iwp.WorkerPool.GetWorker(opts)
+	if err != nil {
+		return w, err
+	}
+
+	return iwp.Wrapper(w), nil
+}
```
tjholm, line 33: I think we could look at adding context to the Handle methods in the worker interface and the wrap workers ass we Add them to the Instrumented worker pool rather than when we retrieve them.

pkg/span/span.go
```diff
@@ -0,0 +1,78 @@
...
+func init() {
+	// support for aws, gcp and azure
+	for _, en := range []string{"AWS_LAMBDA_FUNCTION_NAME", "K_SERVICE", "WEBSITE_SITE_NAME"} {
+		name := os.Getenv(en)
+		if name != "" {
+			FunctionName = name
+			break
+		}
+	}
+}
```
tjholm, line 43: Could we place this support with the relevant provider code to avoid cross-polluting our cloud implementations?

pkg/adapters/grpc/queue_grpc.go
```diff
@@ -62,7 +65,21 @@ func (s *QueueServiceServer) Send(ctx context.Context, req *pb.QueueSendRequest)
...
+		semconv.MessagingDestinationKindQueue,
+		semconv.MessagingDestinationKey.String(req.GetQueue()),
+		semconv.MessagingMessageIDKey.String(nitricTask.ID),
+		attribute.Key("messaging.message_type").String(nitricTask.PayloadType),
+	)
+
+	defer sn.End()
+
 	if err := s.plugin.Send(req.GetQueue(), nitricTask); err != nil {
+		sn.RecordError(err)
```
tjholm, line 81: I'm assuming it's fine for this to run unconditionally?

tjholm (changes requested):
Looking good, just a few questions and comments, planning to go over this a few more times as well.

pkg/adapters/grpc/queue_grpc.go
asalkeld: yes, if you haven't set a traceProvider you get a noop trace provider

pkg/pm/processmanager.go
asalkeld: they are started at different times.. I'll see what I can do.

pkg/worker/instrumented_pool.go
tjholm: Could we update this do perform the wrapping when we add the workers instead?

```go
// AddWorker - Adds the given worker to this pool
func (iwp *InstrumentedWorkerPool) AddWorker(w Worker) error {
        return iwp.WorkerPool.AddWorker(iwp.Wrapper(w))
}
```

asalkeld: ok

pkg/membrane/membrane.go
```diff
@@ -345,11 +358,34 @@ func New(options *MembraneOptions) (*Membrane, error) {
 		})
 	}
 
+	bin := "/usr/bin/otelcol-contrib"
+	config := "/etc/otelcol/config.yaml"
```
tjholm, line 362: We can probably set these as env variables in the wrapping docker file right?

tjholm (changes requested):

tjholm:
Update on this one, after discussion with @asalkeld looking at native gRPC telemetry to allow us to use native headers (at least in the case on unary calls, for preserving trace information.)

pkg/adapters/grpc/event_grpc.go
```diff
@@ -58,11 +61,25 @@ func (s *EventServiceServer) Publish(ctx context.Context, req *pb.EventPublishRe
 		PayloadType: req.GetEvent().GetPayloadType(),
 		Payload:     req.GetEvent().GetPayload().AsMap(),
 	}
+
+	sp := span.FromHeaders(ctx, "topic-"+req.GetTopic(), map[string][]string{})
+	sp.SetAttributes(
+		semconv.CodeFunctionKey.String("EventService.Publish"),
```
tjholm, line 67: Some functions keys are inconsistent:

Here we specify `EventService.<method>`

But in the Queue service we use `Queue.<method>`.

PR #345, opened 2022-11-02
feat: add Headers for events (to be used by tracing)

asalkeld:


contracts/proto/faas/v1/faas.proto
```diff
@@ -164,6 +164,7 @@ message TopicTriggerContext {
   // The topic the message was published for
   string topic = 1;
 
+  map<string, string> headers = 2;
```
tjholm, line 167: Was thinking we could add this as a tracing specific concept directly in trigger context (where data and mimeType are). This could be shared with HTTP requests as well, which would allow us to implement our OOTB instrumentation without potentially interfering with any additional instrumentation users may have implemented in their own HTTP context.

tjholm (changes requested):

contracts/proto/faas/v1/faas.proto
asalkeld: I think we REALLY do want it in the membrane and user tracing interacting. Definitively not separated as you lose valuable info of related events. For example us seeing membrane events in the context of cloud tracing is a positive..
I'll look at moving this, but it needs to get injected back into http headers eventually.

asalkeld: I think a better option is to add Headers into the context (same struct and function signature)

tjholm: I think we're saying similar things.

My main concern is where the trace information is being written on the wire, currently its in two separate locations:

1) In the headers of the HttpContext
2) In the headers of the EventContext

This means in both the membrane and the SDK we need to write code to cater for both cases

My suggestion is to move the trace information into a single location, where it can be easily access and understood as part of the protobuf contract, this may require a little bit more work in the membrane, but won't require special case handling in the SDKs (one implementation for all triggers instead of one for HTTP and one for events).

```proto
message TriggerRequest {
  // The data in the trigger
  bytes data = 1;

  // Should we supply a mime type for the data?
  // Or rely on context?
  string mime_type = 2;

  // TraceInformation from the membrane
  TraceContext trace_context = 10;

  // The context of the trigger
  oneof context {
    HttpTriggerContext http = 3;
    TopicTriggerContext topic = 4;
  }
}
```

The advantage of doing it this was is this message could also be re-used in our other messages and contracts, as an example when publishing an event:
```proto
message EventPublishRequest {
  // The name of the topic to publish the event to
  string topic = 1 [(validate.rules).string = {
    pattern:   "^\\w+([.\\-]\\w+)*$",
    max_bytes: 256,
  }];

  // The event to be published
  NitricEvent event = 2 [(validate.rules).message.required = true];

  // An optional delay specified in seconds (minimum 10 seconds)
  uint32 delay = 3 [(validate.rules).uint32 = {ignore_empty: true, gte:10, lte: 2592000}];
  
  // Optional trace_context for propagating trace information on the wire
  // This can be packaged into messages on a provider basis (e.g. in GCP pubsub we could use message attributes)
  TraceContext trace_context = 4;
}
```
Which gives us a well understood data type for propagating trace information.

NOTE: In the pure HTTP implementation (i.e. HTTP workers) it is fine to just use the HTTP headers as we communicate exclusively over HTTP, but the http worker implementation is currently defunct and the SDKs communicate exclusively with gRPC at the moment.

PR #347, opened 2022-11-04
Add timezone for cron schedule requests.

tjholm:
Include a timezone property in ScheduleWorker requests.


contracts/proto/faas/v1/faas.proto
```diff
@@ -86,7 +86,10 @@ message ScheduleRate {
 }
 
 message ScheduleCron {
+  // cron expression to use
   string cron = 1;
+  // locale to pin cron job to
+  string timezone = 2;
```
tjholm, line 92: Was thinking we might want to add this generically for all schedule workers as it could be handy for pinning rates to a specific timezone as well. E.g. `schedule("example").timezone("Australia/Sydney").every("day")`, would mean every date at 0:00 Sydney time instead of every day at 0:00 UTC+0

tjholm:
closing for now, will revisit when implementing cron timezone repinning

PR #350, opened 2022-11-10
Feat: telemetry and event headers

asalkeld:


pkg/plugins/events/sns/sns.go
```diff
@@ -39,16 +41,16 @@ type SnsEventService struct {
 	provider  core.AwsProvider
 }
 
-func (s *SnsEventService) getTopics() (map[string]string, error) {
+func (s *SnsEventService) getTopics(ctx context.Context) (map[string]string, error) {
 	return s.provider.GetResources(core.AwsResource_Topic)
 }
 
-func (s *SnsEventService) getStateMachines() (map[string]string, error) {
+func (s *SnsEventService) getStateMachines(ctx context.Context) (map[string]string, error) {
```
tjholm, line 48: ctx unused?

pkg/plugins/events/sns/sns.go
```diff
@@ -39,16 +41,16 @@ type SnsEventService struct {
 	provider  core.AwsProvider
 }
 
-func (s *SnsEventService) getTopics() (map[string]string, error) {
+func (s *SnsEventService) getTopics(ctx context.Context) (map[string]string, error) {
```
tjholm, line 44: ctx unused?

pkg/plugins/gateway/cloudrun/http.go
```diff
@@ -47,6 +49,9 @@ func middleware(ctx *fasthttp.RequestCtx, pool worker.WorkerPool) bool {
 	if err := json.Unmarshal(bodyBytes, &pubsubEvent); err == nil && pubsubEvent.Subscription != "" {
 		// We have an event from pubsub here...
 
+		topic := pubsubEvent.Message.Attributes["x-nitric-topic"]
+		ctx := span.FromHeaders(context.TODO(), "topic-"+topic, triggers.HttpHeaders(&rc.Request.Header))
```
tjholm, line 53: Does this work? I can't see where otel propagation would be occurring using HTTP headers for pubsub message publishing. What tells pubsub to apply these HTTP headers to the subscription?

pkg/providers/gcp/membrane.go
```diff
@@ -29,14 +42,57 @@ import (
...
+	if err != nil {
+		return nil, errors.WithMessagef(err, "NITRIC_TRACE_SAMPLE_PERCENT should be an int not %s", rate)
+	}
+
+	return sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sdktrace.ParentBased(sdktrace.TraceIDRatioBased(rate))),
+		sdktrace.WithResource(res),
+		sdktrace.WithBatcher(exp),
+	), nil
+}
```
tjholm, line 89: would be good to do this consistently, probably in a separate file like you have done for aws.

tjholm (changes requested):
Some initial thoughts at first glance. Will need to take a few more passes over.

pkg/plugins/events/sns/sns.go
asalkeld: i'll pass this down into the resourcesGet

tjholm (approved):

contracts/proto/faas/v1/faas.proto
```diff
@@ -119,6 +122,9 @@ message TriggerRequest {
   // Or rely on context?
   string mime_type = 2;
 
+  // TraceInformation from the membrane
+  TraceContext trace_context = 10;
```
tjholm, line 126:
```suggestion
  // TraceInformation from the membrane
  // Allows tying traces from external triggers (e.g. HttpRequests)
  // into each event request/response pair of the Bidirectional stream.
  // which cannot be facilitated by OOTB stream interceptors from OTEL.
  TraceContext trace_context = 10;
```

tjholm (approved):

HomelessDinosaur (approved):

PR #357, opened 2022-12-09
refactor: nitric monorepo

tjholm:
Quick overview of the new structure:

core/ - Contains shared/common core runtime and deployment code (this might be further decomposed later)

e2e/ - Contains e2e and integration tests (currently only tests documents)

provider/<provider_name> - Container individual provider modules (each provider is its own go module)"
  - deploy - Contains code that deploys a stack for this provider
  - runtime - Contains code that will be packaged for runtime use (how to integrate execution units with deployed resources)


tjholm:
ignore the sonarcloud check for now.

HomelessDinosaur:
Were you going to rename provider -> cloud or has that been overturned?

tjholm:
> Were you going to rename provider -> cloud or has that been overturned?

worth discussing now I think :)

davemooreuws:
README will need a few updates.

provider/aws/go.mod
```diff
@@ -306,3 +251,7 @@ require (
 	mvdan.cc/lint v0.0.0-20170908181259-adc824a0674b // indirect
 	mvdan.cc/unparam v0.0.0-20220706161116-678bad134442 // indirect
 )
+
+replace github.com/nitrictech/nitric/core => ../../core
```
asalkeld, line 255: if the cli imports the provider, this is not going to work.
basically you need to undo these replaces. One way is to have a PR that just adds core
then a second one that does the remainder of the changes.

provider/aws/go.mod
```diff
@@ -306,3 +251,7 @@ require (
...
+
+replace github.com/nitrictech/nitric/provider/common => ../common
```
asalkeld, line 257: we probably should break common into small focused packages

Makefile
```diff
@@ -0,0 +1,45 @@
+binaries:
```
asalkeld, line 1:
```
providers := aws gcp azure
binaries: $(providers)
    for dir in $(providers); do \
        $(MAKE) -C $$dir; \
    done```

File: provider/aws/go.mod (hunk 1 above)
Comment (tjholm) [reply provider/aws/go.mod:255]: for reproducible builds this becomes a problem without replace as your sum file will always be behind and defeats the point of having a monorepo structure in the first place.

Will see if there is a way we can address this with go workspaces instead

File: go.work [hunk 4]
```diff
@@ -0,0 +1,10 @@
+go 1.19
```
asalkeld, line 1: how does this work? go.work instead of go.mod?

davemooreuws: Think this is part of go workspaces

provider/aws/go.mod
tjholm: Should be able to get this to work with workspaces and remove the replace directives (wasn't working before). Will see if I can resync the dependencies first.

tjholm: replace directives removed and workspace resynced.

provider/aws/go.mod
tjholm: yes, but probably later (at the moment it only contains a single source file :) )

core/Makefile
```diff
@@ -0,0 +1,82 @@
...
+else
+GOLANGCI_LINT_CACHE=${HOME}/.cache/golangci-lint
+endif
+GOLANGCI_LINT ?= GOLANGCI_LINT_CACHE=$(GOLANGCI_LINT_CACHE) go run github.com/golangci/golangci-lint/cmd/golangci-lint
+
+include tools/tools.mk
+
+init: check-gopath go-mod-download install-tools
+	@echo Installing git hooks
+	@find .git/hooks -type l -exec rm {} \; && find .githooks -type f -exec ln -sf ../../{} .git/hooks/ \;
```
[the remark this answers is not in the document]
tjholm: This was in the original makefile. In principal this can be made entirely opt in as well under an `install-hooks/uninstall-hooks` make directive. That way we can leave it to personal preference.

go.work
tjholm: `go.work` is a golang workspace. It's there OOTB monorepo management so we can co-locate our dependencies and have them interoperate in the same git repository.

This helps ensure that these dependencies can be versioned and released together.

https://go.dev/doc/tutorial/workspaces

Makefile
tjholm: have consolidated these now.

go.work
tjholm: There are a few other things we need to do to ensure all dependencies are properly installed external to the go workspace (once this workspace is versioned).

tjholm:
> README will need a few updates.

Have removed quite a bit of it for now, with some other updates.

Want to create more focused documentation for each of the modules instead.

jyecusch:
> > Were you going to rename provider -> cloud or has that been overturned?
> 
> worth discussing now I think :)

I think the tradeoffs are:
- cloud is less accurate, since you could have a target that isn't a cloud provided potentially (i.e. locally hosted services)
- provider is significantly more overloaded and therefore less clear

HomelessDinosaur:
Integration or service could be alternatives to provider.

tjholm:
renamed provider -> cloud

PR #361, opened 2022-12-20
deployment contract

tjholm:
WIP contract work for decoupling deployments from the nitric CLI.

A bit conflicted on how much in common we should share with the existing resource API.

This contract is developed under the assumption that the nitric CLI will still perform all of the collection logic with the application:

i.e. collect -> compile (into the model for this contract) -> deploy (invoke deployment server with compiled bill of materials).


contracts/proto/deploy/v1/deploy.proto
```diff
@@ -0,0 +1,189 @@
...
+// The Nitric Deloyment Service contract
+service DeployService {
+    // Begins a new deployment
+    // Server will stream updates back to the connected client
+    // on the status of the deployment
+    rpc Up (DeployUpRequest) returns (stream DeployUpEvent);
+    // Tears down an existing deployment
+    // Server will stream updates back to the connected client
+    // on the status of the teardown
+    rpc Down (DeployDownRequest) returns (stream DeployDownEvent);
```
asalkeld, line 23: - List
- Preview
- GetConfigSchema // used by `nitric stack new` so we can ask the user for values that are required

tjholm: For **List** it would be good to get an idea of exactly how valuable this is, it also won't apply in all cases.

**Preview** is possible, but also may not apply in all cases, could possibly use `attributes` or similar property in `Up` to accomodate this and allow providers to document how a `dry-run` could be performed.

**GetConfigSchema** is definetely nice QoL enhancement :+1: not sure it needs to block the initial contract changes though. A new feature could be raised for this and stack file formats documented with providers.

asalkeld: The other thing was something along the lines of `SupportedResourceTypes()` so the cli can error if the user
is using a resource that the backend does not support.

asalkeld: this would be handy for the Schedule in Azure as an example..

asalkeld (approved):

contracts/proto/deploy/v1/deploy.proto
tjholm: That makes sense. At the moment a provider could actually just loop over all provided resources and return its own error on Up (like a pre-deploy/dry-run check) as well. But explicit support wouldn't hurt. Might be something we can do with `Preview` as well.

PR #363, opened 2023-01-09
Expose unary service for worker triggers

tjholm:


cloud/aws/runtime/gateway/lambda.go
```diff
@@ -211,65 +228,64 @@ func (s *LambdaGateway) handle(ctx context.Context, data map[string]interface{})
...
+			})
+			if err != nil {
+				return nil, fmt.Errorf("unable to get worker to event trigger")
+			}
+
+			// FIXME: Restore this with contract update for topic/event context
+			// var mc propagation.MapCarrier = event.Attributes
+
+			if _, err := wrkr.HandleTrigger(ctx, request); err != nil {
+				return nil, err
```
tjholm, line 288: Need to fix this to preserve tracing headers/attributes

PR #364, opened 2023-01-16
Standalone AWS deployment server

tjholm:
Implements an AWS deployment engine as a standalone independent unit (decoupled from the nitric CLI).

This can be tested with a gRPC client (until we integrate it with the CLI).

If we port the remaining providers to this method this makes it possible to decouple our CLI from pulumi as well (could deliver pulumi within docker containers instead.

This also allows providers to be pulled as part of a stack definition (rather than being mapped directly to CLI implementations.

Example Stack file:
```yaml
name: my-stack
provider: aws
```

Could potentially turn into something like:
```yaml
name: my-stack
# Define remote engine to pull and execute (this doesn't have to be docker, just an example
# Mainly demonstrating that this opens to door to building completely custom providers, even in different
# languages/deployment tech
provider: docker://nitrictech/aws-cloud
```


jyecusch:
> ```yaml
> name: my-stack
> # Define remote engine to pull and execute (this doesn't have to be docker, just an example
> # Mainly demonstrating that this opens to door to building completely custom provders.
> provider: docker://nitrictech/aws-cloud
> ```

I like this idea for defining and loading providers

tjholm:
> > ```yaml
> > name: my-stack
> > # Define remote engine to pull and execute (this doesn't have to be docker, just an example
> > # Mainly demonstrating that this opens to door to building completely custom provders.
> > provider: docker://nitrictech/aws-cloud
> > ```
> 
> I like this idea for defining and loading providers. If we go ahead with something like this, maybe the current providers (e.g. `aws`) can become aliases of something like `docker://nitrictech/aws-cloud` as a way to enable backwards compatibility?

that sounds perfect! easy to implement and explain.

The part I'm currently wrapping my head around is configuration for mapping to something like docker.

So if we're using something like terraform/pulumi in a docker container, ensure we properly respect locally configured settings (by mapping configuration directories from the host to the container). So if a user is deploying with a local backend their config doesn't get nuked everytime and existing stacks are properly updated.

asalkeld:
> 

Shouldn't we be able to define the provider per resource rather than per stack?
What if you wanted all built in resources, except this one?

tjholm:
> > 
> 
> Shouldn't we be able to define the provider per resource rather than per stack? What if you wanted all built in resources, except this one?

@asalkeld it depends on where control of the specific resources should be.

Also the idea of a single provider isn't mutually exclusive with swappable/pluggable resources, there are no reasons that a provider couldn't give multiple alternatives for a single resource type within the scope of that provider, that provider could even define its own interface for plugins as well and incorporate them into a build/wrapping process that dynamically includes its own plugins on request.

I feel like pulling all the complexity of individually pluggable resources forward into the stack level might just end up introducing an inner platform effect of configuration into stack rather than that configuration being defined within the scope of a concrete provider.

If you have some ideas on how we could achieve that level of pluggability with high cohesion that would be awesome, how do you think something like that would look?

asalkeld (approved):

PR #367, opened 2023-01-19
Azure deploy server

asalkeld:
it compiles :-O


cloud/azure/deploy/containerapp.go
```diff
@@ -0,0 +1,415 @@
...
+var RoleDefinitions = map[string]string{
+	"KVSecretsOfficer":    "b86a8fe4-44ce-4948-aee5-eccb2c155cd7",
+	"BlobDataContrib":     "ba92f5b4-2d11-453d-a403-e96b0029c9fe",
+	"QueueDataContrib":    "974c5e8b-45b9-4653-ba55-5f855dd0fb88",
+	"EventGridDataSender": "d5a91429-5739-47e2-a06b-3470a27159e7",
+	// Access for locating resources
+	"TagContributor": "4a9ae827-6dc8-4573-8ac7-8239d42aa03f",
+}
+
+func newContainerApp(ctx *pulumi.Context, name string, args *ContainerAppArgs, opts ...pulumi.ResourceOption) (*ContainerApp, error) {
```
tjholm, line 259: Could deployment code be grouped under packages for the resource type they service, similiar to the AWS refactor, also with public`new` methods. (So they're easier to re-use some of these in external provider implementations).

tjholm (changes requested):
Looks good so far, still reading ocer it. Might be good tgo decompose some of the resources we're deploying as well, like subscriptions.

cloud/azure/deploy/subscriptions.go
```diff
@@ -0,0 +1,130 @@
...
+	Apps              map[string]*ContainerApp
+}
+
+type Subscriptions struct {
+	pulumi.ResourceState
+
+	Name string
+}
+
+func newSubscriptions(ctx *pulumi.Context, name string, args *SubscriptionsArgs, opts ...pulumi.ResourceOption) (*Subscriptions, error) {
```
tjholm, line 41: Let's decompose this to deploy individual subscriptions

cloud/azure/go.mod
```diff
@@ -3,48 +3,90 @@ module github.com/nitrictech/nitric/cloud/azure
...
+	github.com/Azure/go-autorest/autorest v0.11.28
+	github.com/Azure/go-autorest/autorest/adal v0.9.21
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.8
 	github.com/Azure/go-autorest/autorest/date v0.3.0
+	github.com/getkin/kin-openapi v0.112.0
 	github.com/golang/mock v1.6.0
 	github.com/golangci/golangci-lint v1.50.1
 	github.com/google/addlicense v1.1.0
 	github.com/mitchellh/mapstructure v1.5.0
+	github.com/nitrictech/cli v1.14.3
```
tjholm, line 18: Remove dependency on nitric cli. I think this is mostly the common tags stuff which should now be available in cloud/common.

PR #374, opened 2023-02-07
Add Deployment Code to Nitric monorepo

tjholm:
TODO:

- [x] Test AWS Provider
- [x] Test GCP Provider
- [x] Test Azure Provider
- [x] Validate any more deploy contract changes before final release.

### How to test

To help with testing you can use the convenience make steps available in each of the provider implementations:

`make installdeploy` will install the provider binary for the cloud you're cd'd into the `$HOME/.nitric/providers` directory. This will always be with a version of `0.0.1`.
> You will also need to build a version of the CLI from the `angus/grpc-deploy` branch.

Then you can configure a stack with that providers ID:

e.g.
```yaml
name: test
provider: nitric/aws@0.0.1
region: ap-southeast-2
```

Additional parameters may be required depending on the provider, each provider should provide error information detailng what properties are missing from your stack file. Running into these errors will help in improving the CLI output (found on already testing azure deployments).

`rpc error: code = InvalidArgument desc = org is not set of invalid` (definetely needs to be cleaned up).

Also good to look over:
https://github.com/nitrictech/nitric/blob/feat/framework-refactor/contracts/proto/deploy/v1/deploy.proto

In detail, to see what is missing or anything that could quickly become deprecated.

Old providers will remain in CLI and will be cut-over as aliases of the new nitric providers.

### TLDR Review Changes
 - https://github.com/nitrictech/nitric/blob/feat/framework-refactor/contracts/proto/deploy/v1/deploy.proto
 - The cloud/*/deploy directories (these can be compared with existing CLI providers or tested against them as well) https://github.com/nitrictech/nitric/tree/feat/framework-refactor/cloud
 - Also the github actions for the deployments: https://github.com/nitrictech/nitric/tree/feat/framework-refactor/.github/workflows (membrane deployments no longer happen, just the deploy binaries (they contain the membrane binaries).


tjholm:
Have just updates attributes to use a struct so arbitrary maps can be sent to a provider via a stack file. This should be good to properly review now.

HomelessDinosaur (approved):

davemooreuws (approved):

PR #380, opened 2023-02-20
Feat/set max concurrent streams

davemooreuws:
# Description

Adds the ability to set the max concurrent stream option from grpc. Also increases the max workers default value to 300.

## Type of change
- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/core/pkg/membrane/membrane.go
+++ b/core/pkg/membrane/membrane.go
@@ -157,7 +157,9 @@ func (s *Membrane) Start() error {
 		return err
 	}
 
-	var opts []grpc.ServerOption
+	opts := []grpc.ServerOption{
+		grpc.MaxConcurrentStreams(uint32(s.pool.GetMaxWorkers())),
+	}
 
 	if s.createTracerProvider != nil {
 		tp, err := s.createTracerProvider(context.Background())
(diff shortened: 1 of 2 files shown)


HomelessDinosaur (approved):

core/pkg/membrane/membrane.go
```diff
@@ -348,20 +354,24 @@ func New(options *MembraneOptions) (*Membrane, error) {
...
-		}
+	maxWorkersEnv := utils.GetEnv("MAX_WORKERS", "300")
+	maxWorkers, err := strconv.Atoi(maxWorkersEnv)
+	if err != nil || minWorkers < 0 {
+		return nil, fmt.Errorf("invalid MAX_WORKERS env var, expected non-negative integer value, got %v", maxWorkersEnv)
+	}
+
+	if minWorkers > maxWorkers {
+		maxWorkers = minWorkers
+	}
```
tjholm, line 371:
```go
maxWorkers = math.Max(minWorkers, maxWorkers)
```

core/pkg/membrane/membrane.go
```diff
@@ -115,6 +115,10 @@ type Membrane struct {
 
 	// Worker pool
 	pool worker.WorkerPool
+
+	// min/max workers
+	minWorkers int
+	maxWorkers int
```
tjholm, line 121: can we make these properties readable from pool?

tjholm (changes requested):

core/pkg/membrane/membrane.go
davemooreuws: only works with floats, that is why I did it this way. Could convert as alternative

tjholm (approved):

PR #390, opened 2023-03-15
feat: Execution Unit Type

tjholm:
See here for how this would be implemented in practice.

https://github.com/nitrictech/nitric/pull/404


contracts/proto/deploy/v1/deploy.proto
```diff
@@ -108,6 +108,9 @@ message ExecutionUnit {
     int32 timeout = 11;
     // Configurable memory size for this instance
     int32 memory = 12;
+
+    // Flexible provider specific properties
+    google.protobuf.Struct properties = 13;
```
tjholm, line 113: Do we want to add `target` as its own property, assuming all providers implement multiple target types or leave it to each provider implementation to specify its own properties. One thought I had with properties if left at the provider level is being able to encapsulate unique properties using the `target` in this case as a property key.

I would even deprecating properties like `timeout` and `memory` and applying them

As an example:
```yaml
name: develop
provider: aws
region: ap-southeast-2
config: 
  # default function specs (all functions inherit these properties by default
  default:
    lambda:
      # Adding verbose name here as an example that not all services may share the same syntax for common properties
       memory: 256
       timeout: 15
    lightsail:
       # lighsail container services are configured via a 'power' property
       power: nano
       
    # specify preferred default deployment platform  
    target: lambda
  # # TODO: Need a way to properly target a deployable function, currently names are inferred from files
  # # but dockerfiles should be different to ensure they're re-usable.
  stateful:
    target: lightsail
    lightsail:
       # override lightsail defaults if any (will be merged with above defaults)


```

tjholm: Have gone with a simple `type` property, it will be up to provider developers how much or how little config they want to allow in the stack file.

HomelessDinosaur (approved):

HomelessDinosaur:
Just need to update our providers to use the non-deprecated properties

PR #406, opened 2023-03-28
feat: Azure config

tjholm:
Adds azure config, note `memory` config for azure container apps can only be done in gigabytes, so it will differ from the other cloud memory configs (want the config to be closer to the actual cloud rather than perform arbitrary conversions).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,7 +25,7 @@ jobs:
         path: |
           ~/.cache/go-build
           ~/go/pkg/mod
-        key: ${{ runner.os }}-golang-${{ hashFiles('**/go.sum') }}
+        key: ${{ runner.os }}-golang-${{ hashFiles('**/go.sum', '**/work.sum') }}
         restore-keys: |
           ${{ runner.os }}-golang-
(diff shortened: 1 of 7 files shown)


HomelessDinosaur:
> Adds azure config, note `memory` config for azure container apps can only be done in gigabytes, so it will differ from the other cloud memory configs (want the config to be closer to the actual cloud rather than perform arbitrary conversions).

Should we just use megabytes as a standard across our providers, and then convert to gigabytes in the azure implementation?

tjholm:
> > Adds azure config, note `memory` config for azure container apps can only be done in gigabytes, so it will differ from the other cloud memory configs (want the config to be closer to the actual cloud rather than perform arbitrary conversions).
> 
> Should we just use megabytes as a standard across our providers, and then convert to gigabytes in the azure implementation? Would save someone from possibly allocating 128 GB accidentally.

I think being more native would be better:
https://learn.microsoft.com/en-us/azure/container-apps/containers#configuration

We could also rename the property to imply the unit as well e.g. `memoryGb`

HomelessDinosaur (approved):

PR #418, opened 2023-03-31
ci: Add actions cache to releaser actions

tjholm:


cloud/aws/deploy/down.go
```diff
@@ -28,25 +28,6 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-type DownStreamMessageWriter struct {
```
HomelessDinosaur, line 31: Is deleting this intentional?

cloud/azure/deploy/down.go
```diff
@@ -25,25 +25,6 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-type DownStreamMessageWriter struct {
```
HomelessDinosaur, line 28: Is this intentional?

tjholm: these should be using the common message writer now

cloud/aws/deploy/down.go
tjholm: these should be using the common message writer now

HomelessDinosaur (approved):

.github/workflows/release-prod.yaml
```diff
@@ -17,8 +17,7 @@ jobs:
         id: tag_version
         uses: mathieudutour/github-tag-action@v5.5
         with:
-          # Use GITHUB_TOKEN here to prevent further workflows
-          # generated on 'tag' action
+          dry_run: true
```
davemooreuws, line 20: Just a question, why was this set to dry_run?

davemooreuws (commented):
Looks good to me, just a question on the prod release flow

.github/workflows/release-prod.yaml
tjholm: We want the NITRIC_BOT_TOKEN to push to tag as part of the release, so the tag push action that publishes the binaries get's triggered.

davemooreuws (approved):

PR #419, opened 2023-03-31
feat: add provider for debugging a stack spec

HomelessDinosaur:
Provider for debugging a deployment


cloud/debug/.goreleaser.yaml
```diff
@@ -0,0 +1,38 @@
...
+      {{- else }}{{ tolower .Arch }}{{ end }}
+    format_overrides:
+      - goos: windows
+        format: zip
+checksum:
+  name_template: "aws_checksums.txt"
+snapshot:
+  name_template: "{{ incpatch .Version }}-next"
+changelog:
+  skip: true
```
tjholm, line 38: not sure we need to release it for now, but there a quite a few `aws` references in here that could be dangerous

jyecusch (approved):
Cool idea, might be interesting to add a shorthand to the CLI at some point to run this. Something like `nitric up --debugonly` or just `nitric debug`

tjholm (approved):

Issue #410, opened 2023-03-28
Design: Subscribe to bucket changes


tjholm:
Design a way to subscribe to changes in buckets as part of application code.

tjholm:
Documenting previous proposal:
```typescript
import { bucket } from '@nitric/sdk';

// create a bucket called files
const files = bucket("files");

// using pattern of "<event_type>:<file_pattern>"
// subscribe to write events for all files on the bucket
files.on('write:*', async (ctx) => {
  // do something with the event
});
```

raksiv:
qq's 

Will we support all access permissions 'read', 'write', 'delete'?
What format will the file pattern be?

tjholm:
@raksiv good questions, for permissions I think we should just impute them from the subscription, so regardless of whether or not a function actually has access to read/write to a bucket, it can still know about things that have happended with it (even if it can't get access to the contents).

Need to do some research on patterns based on how we plan to support this for each platform. Initially I was thinking a basic glob(ish) pattern.

tjholm:
Ways we can support this functionality:

AWS:
 - SNS Bucket notifications: https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketnotification/#aws-s3-bucketnotification

GCP:
 - Storage Notifications: https://www.pulumi.com/registry/packages/gcp/api-docs/storage/notification
 - Events Arcs: https://www.pulumi.com/registry/packages/gcp/api-docs/eventarc/trigger/

Azure:
 - Event Grid (setting the appropriate scope): https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketnotification/#aws-s3-bucketnotification

tjholm:
Three things that need to be designed for this:

 - SDK interfaces (core developer experience).
 - Runtime interfaces (how we register workers and route events at runtime).
 - Deployment interfaces (how we express these in the nitric cloud spec, for providers to deploy).

HomelessDinosaur:
https://www.notion.so/nitric/Subscribing-to-resource-notifications-da31f63ac01340938b0a60fed6cd457c

PR #428, opened 2023-04-26
feat: Add bucket notification subscribers

HomelessDinosaur:
Closes: https://github.com/nitrictech/nitric/issues/410
- Adds contract change to support bucket notification workers at both runtime and deployment
- Updates proto compiler from v3.19.1 to v3.21.4 to support M1
- Adds bucket notification deployments for each cloud
- Updates GCP and Azure gateway routes to handle bucket notifications at runtime
- Updates Lambda gateway to also support S3 events


contracts/proto/faas/v1/faas.proto
```diff
@@ -89,6 +89,23 @@ message ScheduleCron {
   string cron = 1;
 }
 
+// Notification Workers
+enum EventType {
```
tjholm, line 93: Better to be more specific with the naming of these enum e.g. `BucketNotificationType`

contracts/proto/faas/v1/faas.proto
```diff
@@ -89,6 +89,23 @@ message ScheduleCron {
...
+  All = 0;
+  Created = 1;
+  Deleted = 2;
+}
+
+message BucketNotificationWorker {
+  string bucket = 1;
```
tjholm, line 100: wierd indentation here

contracts/proto/faas/v1/faas.proto
```diff
@@ -89,6 +89,23 @@ message ScheduleCron {
...
+	BucketNotificationConfig config = 2;
+}
+
+message BucketNotificationConfig {
+	EventType eventType = 1;
+	string eventFilter = 2;
```
tjholm, line 106: change to `event_filter`.

Also Can you add a quick doc comment on what an eventFilter looks like

contracts/proto/faas/v1/faas.proto
```diff
@@ -103,6 +120,7 @@ message InitRequest {
     ApiWorker api = 10;
     SubscriptionWorker subscription = 11;
     ScheduleWorker schedule = 12;
+    BucketNotificationWorker bucketNotification = 13;
```
tjholm, line 123: `bucket_notification`

contracts/proto/faas/v1/faas.proto
```diff
@@ -176,6 +197,17 @@ message TopicTriggerContext {
...
+enum NotificationType {
+  Bucket = 0;
+}
+
+message NotificationTriggerContext {
+  // The resource the message was triggered by
+  NotificationType type = 1;
+  string resource = 2;
+  map<string, string> attributes = 3; // Convention for buckets is "key" and "type"
+}
```
tjholm, line 209: If the intention here is to future proof, I would avoid including conventions specific to buckets in the base context.

It easier to use oneof in this case to get the specific type with common attributes in the base context.

e.g.

```protobuf
message BucketNotification {
  map<string, string> attributes = 2;
}

message NotificationTriggerContext {
  string source = 1;
  
  // Now we don't need to specify a type enum, the specific contents/type comes in a single field of the context.
  oneof notification {
     BucketNotification bucket = 10;
  }
}
```

cloud/gcp/deploy/up.go
```diff
@@ -266,6 +250,41 @@ func (d *DeployServer) Up(request *deploy.DeployUpRequest, stream deploy.DeployS
...
+						Location: details.Region,
+						Bucket:   buckets[res.Name],
+						Config:   notification.Config,
+						Function: unit,
+					}, defaultResourceOptions)
+					if err != nil {
+						return err
+					}
+				}
+			}
```
tjholm, line 285: Is this safe? Can we guarantee that the order of notifications is deterministic?

Probably best to include more specific unique information, like the target name instead of the index.

tjholm (dismissed):

cloud/aws/deploy/up.go
```diff
@@ -237,6 +221,35 @@ func (d *DeployServer) Up(request *deploy.DeployUpRequest, stream deploy.DeployS
...
+						StackID:      stackID,
+						Location:     details.Region,
+						Bucket:       buckets[res.Name],
+						Functions:    execs,
+						Notification: b.Bucket.Notifications,
+					})
+					if err != nil {
+						return err
+					}
+				}
```
tjholm, line 249: Is is simpler to deploy all notifications at once? Did we want to separate these out by target like you have for GCP

cloud/azure/deploy/up.go
```diff
@@ -284,6 +272,35 @@ func (d *DeployServer) Up(request *deploy.DeployUpRequest, stream deploy.DeployS
...
+				_, err := bucket.NewAzureBucketNotification(ctx, fmt.Sprintf("%s-%d-notification", b.Name, idx), &bucket.AzureBucketNotificationArgs{
+					Bucket:         azBucket,
+					StorageAccount: storageAccount,
+					Config:         notification.Config,
+					Target:         unit,
+				})
+				if err != nil {
+					return err
+				}
+			}
```
tjholm, line 301: Same as GCP not sure that the use of idx for unique naming here is safe.

cloud/aws/deploy/up.go
HomelessDinosaur: Pulumi only supports deploying one S3 Notification per bucket.

tjholm (approved):

davemooreuws (approved):

PR #440, opened 2023-05-08
feat: add declare hook to resource plugin

davemooreuws:
# Description

Adds declare to the resource plugin so we can gather the resources during nitric start in CLI.

### DIFF
--- a/cloud/aws/runtime/core/provider.go
+++ b/cloud/aws/runtime/core/provider.go
@@ -66,6 +66,10 @@ type awsProviderImpl struct {
 
 var _ AwsProvider = &awsProviderImpl{}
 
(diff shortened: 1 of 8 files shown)


cloud/aws/runtime/core/provider.go
```diff
@@ -66,6 +66,10 @@ type awsProviderImpl struct {
 
 var _ AwsProvider = &awsProviderImpl{}
 
+func (a *awsProviderImpl) Declare(ctx context.Context, typ common.ResourceType, name string) error {
+	return nil
+}
```
tjholm, line 71: Might also want to include additional resource details from the payload to save us have to re-implement this later if we're missing information.

tjholm (changes requested):

tjholm (approved):

HomelessDinosaur (approved):

PR #446, opened 2023-05-23
feat: add error if worker type is incompatible with CLI version

HomelessDinosaur:


core/pkg/adapters/grpc/faas_grpc.go
```diff
@@ -72,8 +73,7 @@ func (s *FaasServer) TriggerStream(stream pb.FaasService_TriggerStreamServer) er
 			Notification: notification,
 		})
 	} else {
-		// XXX: Catch all worker type
-		wrkr = worker.NewFaasWorker(adapter)
+		return fmt.Errorf("unable to register worker due to incompatible worker type... you may need to upgrade your CLI version")
```
jyecusch, line 76:
```suggestion
		return fmt.Errorf("unable to register incompatible worker. This can be caused by using older nitric CLI versions, an upgrade may resolve this issue.")
```

jyecusch (changes requested):

core/pkg/adapters/grpc/faas_grpc.go
jyecusch: Suggested change to the wording - since the first part may not mean much to people, I cut that short and jumped into the possible cause/solution.

tjholm:
Lets prioritise failing at deploy time first, in our CLI codeconfig collection.

PR #449, opened 2023-05-24
feat(aws): allow existing secret import

tjholm:
Example config

```yaml
name: test-aws
provider: nitric/aws@0.0.1
region: ap-southeast-2
import:
  secrets:
    test: <your_secrets_arn>
```

Behaviours:

 - If you add an import over an existing nitric secret, that secret will be deleted and replaced by the imported secret.
 - If you remove an imported secret that secret will not be deleted, but a new nitric secret will be created to take its place.
 - Imported secrets will be tagged as part of the deployment process. With essential nitric tags
 - If an imported secret cannot be located, deployment will fail.

### DIFF
--- a/cloud/aws/runtime/secret/secrets_manager.go
+++ b/cloud/aws/runtime/secret/secrets_manager.go
@@ -153,6 +153,12 @@ func (s *secretsManagerSecretService) Access(ctx context.Context, sv *secret.Sec
 		)
 	}
 
+	returnValue := result.SecretBinary
+
+	if returnValue == nil && result.SecretString != nil {
+		returnValue = []byte(*result.SecretString)
+	}
+
 	return &secret.SecretAccessResponse{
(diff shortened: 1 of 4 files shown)


cloud/aws/runtime/secret/secrets_manager.go
```diff
@@ -153,14 +153,20 @@ func (s *secretsManagerSecretService) Access(ctx context.Context, sv *secret.Sec
 		)
 	}
 
+	returnValue := result.SecretBinary
+
+	if returnValue == nil && result.SecretString != nil {
+		returnValue = []byte(*result.SecretString)
+	}
+
```
tjholm, line 161: Needed this fix to support other secret data types that were used outside of the scope of the nitric API.

HomelessDinosaur (approved):

davemooreuws (approved):

PR #450, opened 2023-05-24
feat: Improve resource discovery

tjholm:
Update resource discovery to make way for existing resource imports.

This will allow nitric stacks to share common resources.

Example of new naming convention in practice:
image
> The convention is `x-nitric-stack-<unique-stack-name>`: `resource_name`

This essentially allows a resource to have a different nitric name for a number of different stacks


jyecusch:
Replaced by https://github.com/nitrictech/nitric/pull/514

PR #462, opened 2023-06-07
fix: Add refresh flag to providers

tjholm:
Setting this flag to true will get stacks to refresh from existing infrastructure

e.g.

```yaml
name: test-aws
provider: nitric/aws@0.0.1
region: ap-southeast-2
refresh: true
```


HomelessDinosaur (approved):

davemooreuws (approved):
Looks good, just a lint error.

jyecusch (approved):

PR #528, opened 2023-10-24
fix: return PermissionDenied for resource permission errors in GCP

jyecusch:


tjholm (approved):

HomelessDinosaur:
If a user doesn't have permission to list resources and their labels (like buckets and topics) the error is wrapped when returned, so it won't be picked up as a PermissionDenied error (see [here](https://github.com/nitrictech/nitric/blob/develop/cloud/gcp/runtime/events/pubsub.go#L67C35-L67C35)). It shouldn't be a problem for now since every function currently has list permissions, just something to think about.

HomelessDinosaur (approved):

raksiv (approved):

PR #538, opened 2023-11-07
fix: remove policy document hashing to enable inline updates.

tjholm:
Currently Inline updates are challenging for policies as they contain the resources related the the update in the assignment name

### DIFF
--- a/cloud/aws/deploy/policy/iam.go
+++ b/cloud/aws/deploy/policy/iam.go
@@ -17,8 +17,8 @@
 package policy
 
 import (
-	"crypto/md5" //#nosec G501 -- md5 used only to produce a unique ID from non-sensistive information (policy IDs)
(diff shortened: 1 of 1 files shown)


davemooreuws (approved):

cloud/aws/deploy/policy/iam.go
```diff
@@ -17,8 +17,8 @@
 package policy
 
 import (
-	"crypto/md5" //#nosec G501 -- md5 used only to produce a unique ID from non-sensistive information (policy IDs)
-	"encoding/hex"
+	//#nosec G501 -- md5 used only to produce a unique ID from non-sensistive information (policy IDs)
```
HomelessDinosaur, line 20: Was this nosec on the package that was removed?

PR #539, opened 2023-11-07
Feat/cors

### LINKED ISSUE #520: CORS helpers
While CORS can currently be handled using middleware and by registering OPTIONS routes, we'd like to build this support into the framework to make it more convinient and ideally handle the responses as high up the infra stack as possible. For example, providing support for static responses to OPTIONS requests at the API Gateway or Nitric Server layers to reduce latency and hosting costs.

This support will be needed in all language SDKs.

davemooreuws:
Adding cors support for AWS, Azure, GCP and local development.

closes #520


davemooreuws:
Moving to draft, this will be added post v1

PR #543, opened 2023-11-22
feat: add preview to providers and custom tea lifecycle

HomelessDinosaur:


davemooreuws:
Tested on windows, CLI version 1.33.1 with the AWS provider.

I am still getting that old windows delay. But I am also getting a mass duplication of the screen being printed multiple times.

image

PR #549, opened 2024-01-10
proposal(v1): Documents -> KV

tjholm:
A proposal for simplifying documents to a simple KV store.

With the intention to look at implementing RDS deployment support for richer datastores and querying capability.


nitric/proto/keyvalue/v1/keyvalue.proto
```diff
@@ -0,0 +1,77 @@
...
+}
+
+message KeyValueSetResponse {}
+
+message KeyValueDeleteRequest {
+  // Key of the document to delete
+  Key key = 1;
+}
+
+message KeyValueDeleteResponse {}
```
tjholm, line 77: Highlighting this for visibility, this was a quick rename change, but any structure we want to change here can be done quickly.

PR #550, opened 2024-01-10
fix: issue causing missing api gateway tags on first deploy

jyecusch:
Co-authored-by: Tim Holm <tim.holm@nitric.io>


davemooreuws (commented):
The change looks good. However for the v1 changes in general:
- Looks like we need to bump to go 1.21 for the slices package. Other than that, some linting updates across the code base.

Could fix in this PR as a chore

tjholm (approved):

PR #563, opened 2024-02-04
common logger

jyecusch:
Introduces a common logger, which wraps the standard "log" package. Allows implementing providers to control log level, e.g. disabling the debug output for the local provider in the CLI. We can also add a standard mechanism for specifying the log level for the core providers.


cloud/common/runtime/logger/logger.go
```diff
@@ -0,0 +1,85 @@
...
+	Error(fmt.Sprintf(format, v...))
+}
+
+func Fatal(msg string) {
+	logMsg(FATAL, msg)
+}
+
+func Fatalf(format string, v ...interface{}) {
+	Fatal(fmt.Sprintf(format, v...))
+}
```
tjholm, line 85: Should put this in core instead of runtime common? This introduces a dependency for core on cloud common, but cloud common already depends on core.

davemooreuws (approved):
Looks good, just needs a rebase

PR #590, opened 2024-02-27
fix: handle wildcard chars `*` in azure bucket event filters

jyecusch:


jyecusch:
Also improves the logger by allowing the log level to be set using an environment variable

tjholm (approved):

HomelessDinosaur (approved):

PR #592, opened 2024-02-27
feat: allow listing keys in kv stores

jyecusch:
Also supports key prefix filtering


nitric/proto/kvstore/v1/kvstore.proto
```diff
@@ -0,0 +1,93 @@
...
+option java_package = "io.nitric.proto.kvstore.v1";
+option java_multiple_files = true;
+option java_outer_classname = "KvStore";
+option php_namespace = "Nitric\\Proto\\KvStore\\V1";
+option csharp_namespace = "Nitric.Proto.KvStore.v1";
+
+// Service for storage and retrieval of simple JSON keyValue
+service KvStore {
+  // Get an existing value
+  rpc GetKey (KvStoreGetRequest) returns (KvStoreGetResponse);
```
tjholm, line 18: @jyecusch should we update these to prefix the method name to the messages?
e.g. `KvStoreGetKeysRequest`

tjholm (approved):

PR #594, opened 2024-02-28
fix: port open check for http

davemooreuws:


jyecusch (approved):

core/pkg/workers/http/http.go
```diff
@@ -38,14 +38,35 @@ type HttpRequestHandler interface {
...
-	if err != nil {
-		return false
-	}
+func IsPortOpenWithRetry(host string, connectionTimeout time.Duration, maxRetries int) error {
+	// Create a channel for signaling when the connection is successful or max retries are reached
+	done := make(chan error, 1)
+	defer close(done)
+
+	// Create a ticker to periodically attempt the connection
+	ticker := time.NewTicker(time.Second) // Adjust the interval as needed
```
tjholm, line 47: do we want this to be a bit tighter on timing?

davemooreuws: reduce to 50ms

HomelessDinosaur (approved):

PR #606, opened 2024-04-22
feat: RDS Support

tjholm:
WIP Proposal for SQL Database support as a nitric resource.


davemooreuws:
@tjholm 

Wondering if we need more than just the connection string for libraries that don't support it? Such as a DSN string with gorm etc.
Perhaps add all the database connection variables so that the end user could format and build a string with them?

tjholm:
> @tjholm
> 
> Wondering if we need more than just the connection string for libraries that don't support it? Such as a DSN string with gorm etc. Perhaps add all the database connection variables so that the end user could format and build a string with them?

Its a possiblity, to address GORM as a specific use case, it looks like standard connection strings are also supported (it used pgx parsing under the hood: https://github.com/jackc/pgx/blob/60a01d044a5b3f65b9eea866954fdeea1e7d3f00/pgconn/config.go#L233-L248).

PR #626, opened 2024-06-14
feat(awstf): Add SQL Database support 

tjholm:
Add SQL database deployment support the the AWS terraform provider.


davemooreuws (changes requested):
Does not trigger the code build jobs after 2nd try.

Steps:
Deploy using prisma or drizzle.
Add a 2nd migration
Deploy again

davemooreuws (changes requested):
First deploy works well.

But next ones don't apply the migrations at the correct time, the cloud run builds are triggered before the new image is available.

Also might need to include similar dependencies for deploying services after migrations, like these: https://github.com/nitrictech/nitric/pull/627

jyecusch (approved):

PR #628, opened 2024-06-17
set bash so it works on windows and add region for codebuild

davemooreuws:


cloud/aws/deploytf/.nitric/modules/sql/variables.tf
```diff
@@ -48,6 +48,11 @@ variable "codebuild_role_arn" {
   type        = string
 }
 
+variable "codebuild_region" {
+  description = "The region of the codebuild project"
+  type        = string
+}
```
tjholm, line 54: this needs to be supplied in the code invoking the module as well

PR #648, opened 2024-07-22
feat(azure): Resource Group Imports

tjholm:
WIP

### DIFF
--- a/cloud/azure/deploy/config.go
+++ b/cloud/azure/deploy/config.go
@@ -34,10 +34,16 @@ type AzureContainerAppsConfig struct {
 	MaxReplicas int `mapstructure:"max-replicas"`
 }
 
+type AzImports struct {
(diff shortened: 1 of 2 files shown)


cloud/azure/deploy/config.go
```diff
@@ -34,10 +34,16 @@ type AzureContainerAppsConfig struct {
 	MaxReplicas int `mapstructure:"max-replicas"`
 }
 
+type AzImports struct {
+	// An existing resource group to deploy into
+	ResourceGroup string `mapstructure:"resource-group"`
+}
```
tjholm, line 40: Not sure if we should move this to the top level of config as a simple `resource-group` property as well. In some ways it feels less like its being imported for consumption and more of a configuration directive.

jyecusch (approved):

PR #666, opened 2024-08-20
feat: Batch

tjholm:
Nitric Batch Job support


nitric/proto/batch/v1/batch.proto
```diff
@@ -0,0 +1,44 @@
...
+    // The amount of memory to allocate for the job
+    int64 memory = 2;
+    // The number of GPUs to allocate for the job
+    int64 gpus = 3;
+}
+
+// ClientMessages are sent from the service to the nitric server
+message CreateJobRequest {
+    // The name of the job to create
+    string name = 1;
```
davemooreuws, line 32: Just a question. Is this repeated in the requirements?

nitric/proto/deployments/v1/deployments.proto
```diff
@@ -161,6 +161,31 @@ message Service {
   map<string, string> env = 14; 
 }
 
+message Job {
+  // The name of the job to create
+  string name = 1;
```
davemooreuws, line 166: Do we need job queue name here as well? Or are we handling that.

tjholm: This is deploy time, this is like naming a topic then publishing to that topic using it's name.

nitric/proto/batch/v1/batch.proto
tjholm: That is deploy time, this is runtime. Kind of like how we use the name of a topic to publish to it while also giving it a resource name.

core/pkg/server/job/server.go
```diff
@@ -0,0 +1,96 @@
...
+		websocketServer: websocketspb.UnimplementedWebsocketServer{},
+	}
+
+	for _, option := range options {
+		// Apply the option to the membrane
+		option(membrane)
+	}
+
+	return membrane
+}
```
tjholm, line 96: This duplicates the existing nitric server as a one-off running server with no `Stop` and no `Worker` servers. Can probably de-duplicate between the two here, by specifying a common group of `runtime` plugins that are embedded into each of the Servers.

core/pkg/server/job/server.go
```diff
@@ -0,0 +1,116 @@
...
+	secretspb "github.com/nitrictech/nitric/core/pkg/proto/secrets/v1"
+	sqlpb "github.com/nitrictech/nitric/core/pkg/proto/sql/v1"
+	storagepb "github.com/nitrictech/nitric/core/pkg/proto/storage/v1"
+	topicspb "github.com/nitrictech/nitric/core/pkg/proto/topics/v1"
+	websocketspb "github.com/nitrictech/nitric/core/pkg/proto/websockets/v1"
+	"google.golang.org/grpc"
+)
+
+// NitricJobServer is a membrane for job
+// Created as a sepearate membrane type to avoid overloading the service membrane
```
HomelessDinosaur, line 37:
```suggestion
// Created as a separate membrane type to avoid overloading the service membrane
```

nitric/proto/resources/v1/resources.proto
```diff
@@ -75,6 +79,19 @@ message KeyValueStoreResource {
 message SecretResource {
 }
 
+message JobResourceRequirements {
+  // The number of CPUs to allocate for the job
+  int64 cpus = 1;
+  // The amount of memory to allocate for the job
+  int64 memory = 2;
+  // The number of GPUs to allocate for the job
+  int64 gpus = 3;
```
HomelessDinosaur, line 88: Should these be `double`?

tjholm: That's a good question actually. For values that might be fractional e.g. CPUs possibly. For memory and GPU probably not so much (I haven't seen timesliced provisioned config of either).

Might be worth going over each of the configs to figure that out.

PR #703, opened 2024-12-05
fix(awstf): support SQL deployments without migrations using Terraform

jyecusch:
Allows SQL databases to be deployed using the awstf provider, without configuring database migrations.


davemooreuws:
Tested with prisma, there are some issues:

1. Works the first deployment.
2. Second deployment the code build job migration does not trigger again

jyecusch:
> Tested with prisma, there are some issues:
> 
> 1. Works the first deployment.
> 2. Second deployment the code build job migration does not trigger again

Did that already occur before this change or is it a new issue?

davemooreuws:
> > Tested with prisma, there are some issues:
> > 
> > 1. Works the first deployment.
> > 2. Second deployment the code build job migration does not trigger again
> 
> Did that already occur before this change or is it a new issue?

This PR only, it isn't detecting a change with the image id.

tjholm (approved):

jyecusch:
> > > Tested with prisma, there are some issues:
> > > 
> > > 1. Works the first deployment.
> > > 2. Second deployment the code build job migration does not trigger again
> > 
> > 
> > Did that already occur before this change or is it a new issue?
> 
> Looks like it occurred before, it isn't detecting a change with the image id.

Makes sense, we'll look at a separate fix for that (probably today)

PR #706, opened 2024-12-09
fix(aws): improve the error if an undefined service is declared in APIs

jyecusch:
This error can't occur with standard nitric providers, but catching it helps with custom providers that may edit the API spec before generating the AWS specific spec.

Currently this panics with a invalid memory address or nil pointer dereference error message, which makes it unclear what went wrong.

### DIFF
--- a/cloud/aws/deploytf/api.go
+++ b/cloud/aws/deploytf/api.go
@@ -61,9 +61,6 @@ func awsOperation(op *openapi3.Operation, funcs map[string]*string) *openapi3.Op
 }
 
 func (n *NitricAwsTerraformProvider) Api(stack cdktf.TerraformStack, name string, config *deploymentspb.Api) error {
-	nameArnPairs := map[string]*string{}
-	targetNames := map[string]*string{}
-
 	if config.GetOpenapi() == "" {
 		return fmt.Errorf("aws provider can only deploy OpenAPI specs")
 	}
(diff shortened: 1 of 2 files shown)


tjholm (approved):

cloud/aws/deploytf/api.go
```diff
@@ -99,17 +96,26 @@ func (n *NitricAwsTerraformProvider) Api(stack cdktf.TerraformStack, name string
...
 	nitricServiceTargets := map[string]service.Service{}
-	for _, p := range openapiDoc.Paths {
-		for _, op := range p.Operations() {
-			if v, ok := op.Extensions["x-nitric-target"]; ok {
-				if targetMap, isMap := v.(map[string]any); isMap {
+	for _, apiPath := range openapiDoc.Paths {
+		for _, pathOperation := range apiPath.Operations() {
+			if apiNitricTarget, ok := pathOperation.Extensions["x-nitric-target"]; ok {
+				if targetMap, isMap := apiNitricTarget.(map[string]any); isMap {
 					serviceName := targetMap["name"].(string)
```
raksiv, line 103:
```suggestion
					serviceName, ok := targetMap["name"].(string)
					if !ok {
					    return fmt.Errorf("missing or invalid 'name' field in x-nitric-target for path %s on API 				%s", pathOperation.OperationID, name)
					}
```

davemooreuws (approved):

cloud/aws/deploytf/api.go
raksiv: Probably not reachable with the typical Nitric flow as well, but doesn't hurt :)

PR #725, opened 2025-01-30
feat: deploy static websites

### LINKED ISSUE #728: Deploy static websites AWS and Azure
### Suggestion
Deploy static websites under a CDN with easy access to all APIs on same origin.

- Deploy Storage bucket for website
- Upload all website files
- Deploy CDN 
- Rewrite to apis via `/api/{api_name}/*`
- Invalidate the CDN cache for changed files
- Ability to add sub sites under a path (example /docs)
- Target AWS and Azure, then GCP later

### Value
Ability to deploy websites as well as host APIs under same origin (avoiding CORs setup issues)

davemooreuws:
Closes #728


nitric/proto/deployments/v1/deployments.proto
```diff
@@ -268,6 +268,20 @@ message WebsocketTarget {
...
+  // The index document for the website
+  string index_document = 2;
+
+  // The error document for the website
+  string error_document = 3;
+
+  // The base path for the website
+  string base_path = 4;
+}
+
```
tjholm, line 284: This looks good, the only thing that might be a bit of a sticking point is the ability to access these assets remotely if necessary. We have this issue in other areas of the protobuf spec as well but need to think about how this limits us in being able to host a nitric provider as a dedicated server vs. having to always run it locally. Might be worth including a `oneof` for the actual assets themselves so we have the option of including different asset type references in future.

tjholm (changes requested):

cloud/aws/deploy/website.go
```diff
@@ -0,0 +1,396 @@
...
+	tags := common.Tags(a.StackId, name, resources.Bucket)
+
+	a.publicWebsiteBucket, err = s3.NewBucket(ctx, name, &s3.BucketArgs{
+		Tags: pulumi.ToStringMap(tags),
+	})
+	if err != nil {
+		return err
+	}
+
+	a.Buckets[name] = a.publicWebsiteBucket
```
jyecusch, line 57: This name could conflict with another bucket with the same name in the project declared the usual way.

jyecusch: Same with the tags, I wonder if this needs a different resource type or to not be stored this way at all.

cloud/aws/deploy/website.go
```diff
@@ -0,0 +1,396 @@
...
+	// Enumerate the public directory in pwd and upload all files to the public bucket
+	// This will be the source for our cloudfront distribution
+	err := filepath.WalkDir(cleanedPath, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		// Skip directories
+		if d.IsDir() {
+			return nil
+		}
```
jyecusch, line 82: Do we need to account for other non-file entities (e.g. symlinks, etc.)?

cloud/aws/deploy/website.go
```diff
@@ -0,0 +1,396 @@
...
+			})
+			if err != nil {
+				// TODO: Could find a better way to check if the object doesn't exist
+				if !strings.Contains(err.Error(), "couldn't find resource") {
+					// Propagate other errors
+					return pulumi.String("").ToStringOutput(), err
+				}
+			}
+
+			obj, err := s3.NewBucketObject(ctx, arn, &s3.BucketObjectArgs{
```
jyecusch, line 116: Could a pulumi resource be used here instead? https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketobject/

jyecusch (changes requested):

cloud/aws/deploy/website.go
davemooreuws: > Could a pulumi resource be used here instead? https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketobject/

When using the pulumi resource, If a bucket object is deleted or does not exist yet it will fail the entire deployment

nitric/proto/deployments/v1/deployments.proto
davemooreuws: Perhaps something like:

```
// The source of the website assets
oneof asset_source {
  string local_directory = 5;
  string remote_url = 6;
}
```

cloud/aws/deploy/website.go
davemooreuws: Updated

cloud/aws/deploy/website.go
davemooreuws: updated to skip non-regular files

cloud/aws/deploy/website.go
tjholm: @davemooreuws I can understand the object being deleted causing issues, but do you have a reproduction for it failing when the object doesn't exist that one sounds like a major bug and there are many pulumi example that just wouldn't work if that was the case.

I think for the deletion case we can accept as users should not be manually removing files from these buckets, lest they break their own site.

davemooreuws: @tjholm I think I can switch to the resource for aws, the users can just refresh if it gets out of sync.
I believe it was azure that had the read error issues using pulumi, had to use the sdk for that.

davemooreuws: Did a refactor to use the s3.HeadObject method to get the Etag. This is more performant and we can correctly check the errors. The pulumi read resource methods don't work for this use case (at the moment).

cloud/aws/deploy/website.go
```diff
@@ -0,0 +1,447 @@
...
+					if newEtag != existingETag {
+						return pulumi.String(objectKey).ToStringOutput()
+					}
+
+					return pulumi.String("").ToStringOutput()
+				}).(pulumi.StringOutput), nil
+			}
+
+			return pulumi.String("").ToStringOutput(), nil
+		}).(pulumi.StringOutput)
```
tjholm, line 189: wrapping applys here might lead to some issues in future with resolution and dependency graph traversal.

Could we do this in two seperate applies?
```go
existingTag := a.publicWebsiteBucket.Bucket.ApplyT(func(bucket string) (string, error) {
   existingETag, err := fileETag(context.TODO(), client, bucket, strings.TrimPrefix(objectKey, "/"))
   if err != nil {
     return "", err
   }
})
```
```go
obj, err := s3.NewBucketObject(ctx, arn, &s3.BucketObjectArgs{
  Bucket:      a.publicWebsiteBucket.Bucket,
  Source:      pulumi.NewFileAsset(path),
  ContentType: pulumi.String(contentType),
  Key:         pulumi.String(objectKey),
}, opts...)
if err != nil {
  return pulumi.String("").ToStringOutput(), err
}
```

```go
// return blank if not changed otherwise return new tag...
pulumi.All(existingTag, obj.Etag, func(args []interface{} { ... }))
```

tjholm (dismissed):

cloud/aws/common/index.go
```diff
@@ -32,27 +32,31 @@ type Topic struct {
...
+	Topics         map[string]Topic          `json:"topics"`
+	KvStores       map[string]AwsResourceArn `json:"kvStores"`
+	Queues         map[string]AwsResourceArn `json:"queues"`
+	Secrets        map[string]AwsResourceArn `json:"secrets"`
+	Apis           map[string]ApiGateway     `json:"apis"`
+	HttpProxies    map[string]ApiGateway     `json:"httpProxies"`
+	Websockets     map[string]ApiGateway     `json:"websockets"`
+	Schedules      map[string]AwsResourceArn `json:"schedules"`
+	Distributions  map[string]AwsResourceArn `json:"distributions"`
+	WebsiteBuckets map[string]AwsResourceArn `json:"websiteBuckets"`
```
tjholm, line 45: Are these being referenced at runtime? Do they need to be included in the index for SSM?

davemooreuws: No they are not referenced at runtime

tjholm: can probably take these out of the resource index for now then, generally only used for things we need to access properties of at runtime

cloud/aws/deploy/website.go
davemooreuws: @tjholm I have pushed and tested the changes for this. I'll now check the azure code.

cloud/aws/common/index.go
davemooreuws: Ok I will remove them

cloud/aws/deploy/deploy.go
```diff
@@ -92,6 +100,7 @@ type NitricAwsPulumiProvider struct {
 	Secrets               map[string]*secretsmanager.Secret
 	Buckets               map[string]*s3.Bucket
 	BucketNotifications   map[string]*s3.BucketNotification
+	Distributions         map[string]*cloudfront.Distribution
```
tjholm, line 103: This should probably just be `*cloudfront.Distribution` rather than a map. I don't see us deploying more than one distribution and even if we do its easy enough to work in and we can properly consider how we'll uniquely identify each distribution at that time.

cloud/aws/deploy/website.go
```diff
@@ -0,0 +1,452 @@
...
+						Forward: pulumi.String("all"),
+					},
+					// Headers: pulumi.ToStringArray([]string{"*"}),
+				},
+				ViewerProtocolPolicy: pulumi.String("https-only"),
+			},
+		)
+	}
+
+	name := "website-cdn"
```
tjholm, line 354: not just a CDN for `websites` its really the single entrypoint for the whole application. I don't think we'll ever really deploy more than one per stack, might be better to name it so it can be associated with the deployed nitric stack.

tjholm (changes requested):

cloud/azure/deploy/website.go
```diff
@@ -0,0 +1,522 @@
...
+	// Pull the hostname out of the storage-account endpoint.
+	originHostname := p.StorageAccount.PrimaryEndpoints.ApplyT(func(endpoints storage.EndpointsResponse) (string, error) {
+		parsed, err := url.Parse(endpoints.Web)
+		if err != nil {
+			return "", err
+		}
+		return parsed.Hostname(), nil
+	}).(pulumi.StringOutput)
+
+	endpointName := fmt.Sprintf("website-endpoint-%s", p.StackId)
```
tjholm, line 334: the `website` part here may not be necessary, Including the stackId here is really good though (should be the same in AWS). possibly something like: `fmt.Sprintf("%s-endpoint", p.StackId)` or the other way around.

cloud/aws/deploy/website.go
davemooreuws: Made some updates for this.

cloud/azure/deploy/website.go
davemooreuws: Made some updates for this.

cloud/aws/deploy/deploy.go
davemooreuws: Made some updates for this.

cloud/azure/deploy/website.go
```diff
@@ -0,0 +1,522 @@
...
+				),
+			}
+
+			ruleOrder++
+
+			return rule.ToDeliveryRuleOutput()
+		}).(cdn.DeliveryRuleOutput)
+
+		deliveryRules = append(deliveryRules, deliveryRuleOutput)
+	}
```
tjholm, line 453: is order here always guaranteed between deployments if we're looping over map keys?

cloud/aws/deploy/website.go
```diff
@@ -0,0 +1,450 @@
...
+					QueryString: pulumi.Bool(true),
+					Cookies: &cloudfront.DistributionOrderedCacheBehaviorForwardedValuesCookiesArgs{
+						Forward: pulumi.String("all"),
+					},
+					// Headers: pulumi.ToStringArray([]string{"*"}),
+				},
+				ViewerProtocolPolicy: pulumi.String("https-only"),
+			},
+		)
+	}
```
tjholm, line 352: same question as Azure on determinism when looping over map keys, could this produce changes to the distribution when none are necessary?

tjholm: For map loop determinism since go 1.23
```go
import (
	"maps"
	"slices"
)
...
for _, k := range slices.Sorted(maps.Keys(m)) {
    v := m[k]
}
```

cloud/azure/deploy/website.go
davemooreuws: changes made

cloud/aws/deploy/website.go
davemooreuws: changes made

cloud/azure/deploy/website.go
```diff
@@ -0,0 +1,530 @@
...
+							Parameters: cdn.UrlRewriteActionParameters{
+								Destination:   "/",
+								SourcePattern: fmt.Sprintf("/api/%s/", name),
+								TypeName:      "DeliveryRuleUrlRewriteActionParameters",
+							},
+						},
+						// TODO add cache control
+					},
+				),
+			}
```
tjholm, line 453: Can we use the `Args` equivalents of the inner fields for these in order to avoid wrapping in another `pulumi.All` here?

As an example:
```suggestion
			rule := &cdn.DeliveryRuleArgs{
				Name:  pulumi.Sprintf("forward_%s", name),
				Order: pulumi.Int(ruleOrder),
				Conditions: pulumi.ToArray(
					[]interface{}{
						cdn.DeliveryRuleUrlPathConditionArgs{
							Name: pulumi.String("UrlPath"),
							Parameters: cdn.UrlPathMatchConditionParametersArgs{
								MatchValues: pulumi.StringArray{
									pulumi.Sprintf("/api/%s", name),
								},
								Transforms: pulumi.StringArray{
									pulumi.String(cdn.TransformLowercase),
								},
								TypeName: pulumi.String("DeliveryRuleUrlPathMatchConditionParameters"),
								Operator: pulumi.String(cdn.OperatorBeginsWith),
							},
						},
					}),
				Actions: pulumi.ToArray(
					[]interface{}{
						cdn.OriginGroupOverrideActionArgs{
							Name: pulumi.String("OriginGroupOverride"),
							Parameters: cdn.OriginGroupOverrideActionParametersArgs{
								OriginGroup: cdn.ResourceReferenceArgs{
									Id: pulumi.String(apiOriginGroupId),
								},
								TypeName: pulumi.String("DeliveryRuleOriginGroupOverrideActionParameters"),
							},
						},
						cdn.UrlRewriteActionArgs{
							Name: pulumi.String("UrlRewrite"),
							Parameters: cdn.UrlRewriteActionParametersArgs{
								Destination:   pulumi.String("/"),
								SourcePattern: pulumi.Sprintf("/api/%s/", name),
								TypeName:      pulumi.String("DeliveryRuleUrlRewriteActionParameters"),
							},
						},
						// TODO add cache control
					},
				),
			}
```

tjholm (changes requested):

cloud/azure/deploy/website.go
davemooreuws: Sure can, updated with the changes.

cloud/azure/deploy/website.go
```diff
@@ -0,0 +1,520 @@
...
+		if ok && config.Website.BasePath == "/" {
+			rootWebsite = config.Website
+			break
+		}
+	}
+
+	if rootWebsite == nil {
+		return fmt.Errorf("no root website configuration found")
+	}
+
```
tjholm, line 63: does this check also exist in the AWS implementation?

davemooreuws: I'll move this to the CLI

tjholm (approved):

jyecusch (approved):

PR #733, opened 2025-02-19
Refactor hardcoded roles

raksiv:
# Description

Include a summary of the change or which issue is fixed. Please also include any relevant context including any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Testing

Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

### DIFF
--- a/cloud/azure/deploytf/.nitric/modules/service/main.tf
+++ b/cloud/azure/deploytf/.nitric/modules/service/main.tf
@@ -43,17 +43,26 @@ data "azuread_client_config" "current" {}
 
 data "azurerm_client_config" "current" {}
 
+variable "role_names" {
+  type = map(string)
+  default = {
+    "KVSecretsOfficer"    = "Key Vault Secrets Officer"
+    "BlobDataContrib"     = "Storage Blob Data Contributor"
+    "QueueDataContrib"    = "Storage Queue Data Contributor"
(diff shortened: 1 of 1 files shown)


cloud/azure/deploytf/.nitric/modules/service/main.tf
```diff
@@ -43,17 +43,26 @@ data "azuread_client_config" "current" {}
 
 data "azurerm_client_config" "current" {}
 
+variable "role_names" {
```
davemooreuws, line 46: should this be under /service/variables.tf?

davemooreuws (changes requested):

cloud/azure/deploytf/.nitric/modules/service/main.tf
raksiv: I'm not entirely sure actually, so far the pattern is that all the logic is in the main.tf file and var.tf is mostly only used by CDKTF for class generation.

davemooreuws:
not required

PR #761, opened 2025-03-19
fix(azure): migrate cdn to frontdoor standard pulumi

davemooreuws:
Migrates from classic cdn to front door standard.

In the review double check things like cache behavior settings.

### DIFF
--- a/cloud/azure/deploy/website.go
+++ b/cloud/azure/deploy/website.go
@@ -475,6 +475,10 @@ func (p *NitricAzurePulumiProvider) deployCDN(ctx *pulumi.Context) error {
 									Id: apiOriginGroup.ID(),
 								},
 							},
+							CacheConfiguration: cdn.CacheConfigurationArgs{
+								CacheBehavior:              pulumi.String(cdn.RuleCacheBehaviorHonorOrigin),
+								QueryStringCachingBehavior: pulumi.String(cdn.AfdQueryStringCachingBehaviorIgnoreQueryString),
+							},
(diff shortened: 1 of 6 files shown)


cloud/azure/deploy/website.go
```diff
@@ -518,3 +587,30 @@ func (p *NitricAzurePulumiProvider) deployCDN(ctx *pulumi.Context) error {
 
 	return nil
 }
+
+func getPrimaryEndpointUrl(storageAccount *storage.StorageAccount) pulumi.StringOutput {
```
tjholm, line 591: getWebEndpointHostName might be a more accurate function name here

cloud/azure/deploy/website.go
```diff
@@ -518,3 +690,30 @@ func (p *NitricAzurePulumiProvider) deployCDN(ctx *pulumi.Context) error {
...
+	// Use character position and value to guarantee uniqueness
+	// This is essentially creating a custom numeric representation
+	for i, char := range name {
+		// Multiply by position+1 to weight characters differently
+		// Use prime number multiplier to reduce collision risk
+		base += int(char) * (i + 1) * 31
+	}
+
+	return base
+}
```
tjholm, line 719: If this method is here to workaround issues with rule order conflicts for provisioning new rules before old rules are deleted make sure its clearly documented, without clear documentation is looks like it shouldn't be here.

tjholm (approved):

jyecusch (approved):

PR #783, opened 2025-04-11
chore: remove several 'todo' comments

jyecusch:
All comments are either no longer relevant or have been raised as separate issues on GH

### DIFF
--- a/cloud/aws/deploy/batch.go
+++ b/cloud/aws/deploy/batch.go
@@ -249,7 +249,6 @@ func (p *NitricAwsPulumiProvider) Batch(ctx *pulumi.Context, parent pulumi.Resou
 	}
 
 	listActions := []string{
-		// TODO: test that all resources still work without these permissions
 		"sns:ListTopics",
 		"sqs:ListQueues",
 		"dynamodb:ListTables",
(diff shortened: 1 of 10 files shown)


cloud/aws/deploy/batch.go
```diff
@@ -249,7 +249,6 @@ func (p *NitricAwsPulumiProvider) Batch(ctx *pulumi.Context, parent pulumi.Resou
 	}
 
 	listActions := []string{
-		// TODO: test that all resources still work without these permissions
```
tjholm, line 252: we might be able to remove some of these permissions since re-working to use ssm

cloud/aws/deploy/service.go
```diff
@@ -125,7 +125,6 @@ func (a *NitricAwsPulumiProvider) Service(ctx *pulumi.Context, parent pulumi.Res
 	}
 
 	listActions := []string{
-		// TODO: test that all resources still work without these permissions
```
tjholm, line 128: same as above comment

cloud/aws/deploytf/service.go
```diff
@@ -54,7 +53,6 @@ func (a *NitricAwsTerraformProvider) Service(stack cdktf.TerraformStack, name st
 		// "NITRIC_AWS_RESOURCE_RESOLVER": jsii.String("tagging"),
 	}
 
-	// TODO: Only apply to requesting services
```
tjholm, line 57: Still relevant as databases currently don't have 'requesting service' metadata, all lambdas get access to the database currently

PR #787, opened 2025-04-15
fix(storage): consistently detect and store content-type when writing files

jyecusch:
Apply the same content-type detection scheme to all supported clouds

### DIFF
--- a/cloud/gcp/runtime/storage/storage.go
+++ b/cloud/gcp/runtime/storage/storage.go
@@ -33,6 +33,7 @@ import (
 	"google.golang.org/api/option"
 	"google.golang.org/grpc/codes"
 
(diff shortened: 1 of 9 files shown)


cloud/gcp/runtime/storage/storage.go
```diff
@@ -153,6 +154,7 @@ func (s *StorageStorageService) Write(ctx context.Context, req *storagePb.Storag
 	}
 
 	writer := bucketHandle.Object(req.Key).NewWriter(ctx)
+	writer.ObjectAttrs().ContentType = content.DetectContentType(req.Key, req.Body)
```
tjholm, line 157: Should we also add inference from the key extension if provided?

tjholm (approved):

davemooreuws (approved):

PR #788, opened 2025-04-15
fix(azure): prevent conflicting topic subscription names

jyecusch:
When topic names are long they result in subscriptions names that must be truncated (to max 24 chars) to meet Azure's name restrictions. The truncation could result in two services subscribing to the same topic ending up with the same name for their subscriptions, causing a conflict for the Event Subscription resources.

### DIFF
--- a/cloud/azure/deploy/topic.go
+++ b/cloud/azure/deploy/topic.go
@@ -59,17 +59,20 @@ func (p *NitricAzurePulumiProvider) newEventGridTopicSubscription(ctx *pulumi.Co
 
 	subName := topicName + "-" + config.GetService()
 
-	subscriptionId, err := random.NewRandomId(ctx, subName+"-ID", &random.RandomIdArgs{
-		Prefix:     pulumi.String("sub"),
-		ByteLength: pulumi.Int(EventSubscriptionRT.MaxLen - 3),
-		Keepers:    pulumi.Map{"subName": pulumi.String(subName)},
+	subscriptionId, err := random.NewRandomString(ctx, subName+"-ID", &random.RandomStringArgs{
+		Length:  pulumi.Int(EventSubscriptionRT.MaxLen),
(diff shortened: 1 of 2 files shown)


davemooreuws (approved):

HomelessDinosaur (approved):

cloud/azure/deploy/topic.go
```diff
@@ -58,8 +59,21 @@ func (p *NitricAzurePulumiProvider) newEventGridTopicSubscription(ctx *pulumi.Co
...
+		Upper:   pulumi.Bool(true),
+		Lower:   pulumi.Bool(true),
+		Special: pulumi.Bool(false),
+		Number:  pulumi.Bool(false),
+	})
+	if err != nil {
+		return err
+	}
+
+	_, err = eventgrid.NewEventSubscription(ctx, subName, &eventgrid.EventSubscriptionArgs{
```
davemooreuws, line 74: the URN will not be unique, replace subName with the unique name

davemooreuws (changes requested):

cloud/azure/deploy/topic.go
jyecusch: I think it should be, since the subName value doesn't need to be truncated in the urn

davemooreuws: this is a separate issue, logged here: https://github.com/nitrictech/nitric/issues/790

tjholm: This is already addressed by the provided subscriptionId (see EventSubscriptionName) is it not? (subscriptionId is a completed random string that uses the original subName for recalculation). The pulumi URN should be unique as there should not be more than one topic subscription per service

davemooreuws: We should throw an error if more than one subscription is added

davemooreuws (approved):

davemooreuws (approved):

PR #793, opened 2025-04-28
fix(azure-tf): adds resource name to policy name

### LINKED ISSUE #792: Role policy assignment names are not unique to the resource name in azure terraform
### Issue

Should be able to define two or more roles in a service, for example. 

Two topic publish roles with different names.

### Expected

Should not error.

davemooreuws:
Ensures uniqueness of policy names by including the resource name in the generated name.

Fixes #792

### DIFF
--- a/cloud/azure/deploytf/policy.go
+++ b/cloud/azure/deploytf/policy.go
@@ -177,7 +177,7 @@ func (a *NitricAzureTerraformProvider) Policy(stack cdktf.TerraformStack, name s
 					return err
 				}
 
-				policy.NewPolicy(stack, jsii.Sprintf("%s-%s", principal.Id.Name, roleName), &policy.PolicyConfig{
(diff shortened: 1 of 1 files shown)


jyecusch (approved):

HomelessDinosaur (approved):

HomelessDinosaur:
Should this check the length of the name to limit it to 64 characters? https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftauthorization

tjholm:
Might want to test the behaviour really quick when migrating these values just to make sure the delete replace order for cases where this still works is fine.

davemooreuws:
> Should this check the length of the name to limit it to 64 characters? https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftauthorization

I think this name is only used as the Terraform resource ID for tracking purposes within Terraform's state, not as the actual Azure resource name.

davemooreuws:
> Might want to test the behaviour really quick when migrating these values just to make sure the delete replace order for cases where this still works is fine.

Tested it with topic publish

PR #794, opened 2025-04-28
fix: keep original auth header if x-forwarded-authorization isn't set

jyecusch:


davemooreuws (approved):

cloud/common/runtime/gateway/http.go
```diff
@@ -71,6 +71,9 @@ func HttpHeadersToMap(rh *fasthttp.RequestHeader) map[string][]string {
 		} else if strings.ToLower(keyString) == "authorization" {
 			// Forward original authorization header
 			headerCopy["X-Platform-Authorization"] = []string{string(val)}
+			if _, exists := headerCopy["Authorization"]; !exists {
```
HomelessDinosaur, line 74: Couldn't this have different outcomes based on the ordering of the incoming headers?

tjholm: Only matters if we don't want x-forwarded-authorization to clobber the provided authorization value. Which we probably do, if we've already preserved it inside that value in the API gateway configuration

tjholm (approved):

PR #795, opened 2025-04-28
fix(azure-tf): use unique stack id instead of stack_name variable

### LINKED ISSUE #791: The stack_id is used incorrectly by tags in azure terraform
### Issue

There is an inconsistency with the use of variable stack_id within the terraform stack tags, this causes resource lookups to fail (such as topics). It also makes it confusing due it being called `stack_name` this needs to be changed.

### Steps

Deploy an app with a topic and an api to trigger it.

### Expected

Should not error with topic not found.

davemooreuws:
fixes #791


HomelessDinosaur:
The outputs still have a `stack_id` and a `stack_name` which is now outputting the same value. Did you mean to remove the `stack_name` from the outputs rather than the variables?

davemooreuws:
> The outputs still have a `stack_id` and a `stack_name` which is now outputting the same value. Did you mean to remove the `stack_name` from the outputs rather than the variables?

Pushed some changes for this. I am now setting the service env variable differently.

cloud/azure/deploytf/.nitric/modules/stack/main.tf
```diff
@@ -8,7 +8,8 @@ resource "random_string" "stack_id" {
 }
 
 locals {
-  stack_name = "${var.stack_name}-${random_string.stack_id.result}"
+  stack_id   = random_string.stack_id.result
+  stack_name = "${var.stack_name}-${local.stack_id}"
```
HomelessDinosaur, line 12: This isn't necessarily an issue with this PR in particular but more a source of confusion in general: `stack_name` is ambiguously named because it both refers to the name of the stack based on the stack file and, in this case, also the stack file name with the random id. It may be better to disambiguate these definitions across the board so that we don't run into similar bugs down the line.

cloud/azure/deploytf/.nitric/modules/stack/main.tf
```diff
@@ -17,11 +18,11 @@ data "azurerm_client_config" "current" {}
 resource "azurerm_resource_group" "resource_group" {
   count = var.resource_group_name == null ? 1 : 0
 
-  name     = "${var.stack_name}-rg${random_string.stack_id.result}"
+  name     = "${var.stack_name}-rg${local.stack_id}"
```
HomelessDinosaur, line 21: This is an example where `stack_name` is confusing for quick reading.

HomelessDinosaur (approved):

jyecusch (approved):

PR #796, opened 2025-04-30
fix(awstf): apply correct resource group tag filter

jyecusch:


cloud/aws/deploytf/.nitric/modules/stack/main.tf
```diff
@@ -5,15 +5,15 @@ resource "random_string" "id" {
 }
 
 resource "aws_resourcegroups_group" "group" {
-  name = "nitric-resource-group-${random_string.id.result}"
+  name = "${var.project_name}-${var.stack_name}-${random_string.id.result}"
```
HomelessDinosaur, line 8: Will this force a recreation of all resources in an existing user's AWS project?

davemooreuws: This will only replace the resource group. Other resources won't change.

davemooreuws (approved):

HomelessDinosaur (approved):

PR #808, opened 2025-06-05
Generates SDK clients from application specs

jyecusch:
This pull request generates SDK clients from application specifications.

Key changes:
- Introduces a new SDK module with generated Go clients for buckets.
- Normalizes resource names for consistent code generation across languages.
- Refactors the SDK generation process to generate each SDK separately.
- Removes redundant dependencies.


cli/pkg/sdk/go.go
```diff
@@ -81,205 +87,31 @@ const clientTemplate = `// Code generated by Nitric SDK generator. DO NOT EDIT.
 package {{.Package}}
 
 import (
```
tjholm, line 89: can we embed the client template from another file here?

tjholm (approved):

PR #810, opened 2025-06-09
Add cloudfront entrypoint plugin

tjholm:


engines/terraform/plugins/awslambda/module/outputs.tf
```diff
@@ -1,7 +1,10 @@
 output "nitric" {
   value = {
     id            = aws_lambda_function.function.arn
-    http_endpoint = aws_lambda_function_url.endpoint.function_url
+    domain_name = split("/", aws_lambda_function_url.endpoint.function_url)[2]
+    raw = {
```
tjholm, line 5: thinking of changing this key to `exports` or even `exports.resources` to allow other export types (e.g. files/env variables)

jyecusch (approved):

PR #833, opened 2025-06-19
Google Cloud CDN, Storage, Cloud Run, and Service Accounts

HomelessDinosaur:


engines/terraform/plugins/gcpiamserviceaccount/module/variables.tf
```diff
@@ -0,0 +1,18 @@
+variable "nitric" {
+  type = object({
+    name = string
+    stack_id = string
```
tjholm, line 4: if not already included we'll need to add this to the `terraform.go` which we can do in a separate PR, but this is a good catch

PR #836, opened 2025-06-21
add cli auth process

jyecusch:


tjholm (changes requested):
remove manifest change

cli/internal/auth/auth.go
```diff
@@ -1,4 +1,164 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = <-p.done
+	if err != nil {
+		fmt.Printf("\n%s Login failed due to an error: %s\n", style.Red(icons.Cross), err)
+	}
+
+	p.callbackServer.Shutdown(context.Background())
```
tjholm, line 125: Will callbackServer be referenced in other contexts in the future or just within this method? Can probably be made stateless if the latter.

tjholm: Also can probably separate out the listener and the server? If so we can make the callback port dynamic if that's possible

tjholm (changes requested):
Does callback server need to be stateful?

Also we can take advantage of net.Listen() port finding capability to get the next available port and make the callback port dynamic. (I think this should work `net.Listen("127.0.0.1:0")`

cli/internal/auth/auth.go
```diff
@@ -1,4 +1,164 @@
...
+)
+
+//go:embed login_success.html
+var loginSuccessPage []byte
+
 type Auth interface {
 }
+
+// TODO: These values are not secret, but we may want to pull them remotely incase of a change.
+var AUTH_SERVER_PORT = 54321
```
davemooreuws, line 25:
```suggestion
var AUTH_SERVER_PORT = 48321
```

cli/internal/auth/auth.go
```diff
@@ -1,4 +1,164 @@
...
+	// Start the Flow
+	var err error
+	p.pkceChallenge, err = workos.CreatePkceChallenge()
+	if err != nil {
+		return err
+	}
+
+	authUrl, err := p.client.GetAuthorizationUrl(workos.GetAuthorizationUrlOptions{
+		Provider:            "authkit",
+		RedirectURI:         "http://127.0.0.1:54321/callback",
```
davemooreuws, line 104:
```suggestion
		RedirectURI:         fmt.Sprintf("http://127.0.0.1:%d/callback", AUTH_SERVER_PORT),
```

davemooreuws (changes requested):
use a port that is not in the restricted port range on windows.

tjholm (approved):

PR #842, opened 2025-06-25
Design file sync

tjholm:


jyecusch:
Looks good, I made a change to how it creates the UI URL (using the local config).

PR #845, opened 2025-06-29
fix: add service intent env variables to simulation

davemooreuws:


davemooreuws:
I will close this and create a new issue for entrypoint port ranges in our backlog for local dev.

PR #847, opened 2025-07-01
Add python sdk

raksiv:
- Generate the python SDK
- Implement Buckets handler

sdk is published to test pypi and pypi under temporary name - 

https://test.pypi.org/project/veetwo-client/0.1.0/
https://pypi.org/project/veetwo-client/0.1.0/

sample project can be seen here - 
https://github.com/raksiv/test-python-v2#

### DIFF
--- a/cli/pkg/client/client.go
+++ b/cli/pkg/client/client.go
@@ -1,12 +1,16 @@
 package client
 
 import (
-	"fmt"
-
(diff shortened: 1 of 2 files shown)


HomelessDinosaur (approved):

raksiv:
Thanks, I've updated the sample project with your suggestions!

PR #849, opened 2025-07-03
add `new` command to new cli

jyecusch:
Closes: NIT-38


cli/pkg/schema/file.go
```diff
@@ -6,9 +6,10 @@ import (
 
 	"github.com/spf13/afero"
 	"github.com/xeipuuv/gojsonschema"
+	"gopkg.in/yaml.v2"
 )
 
-func LoadFromFile(fs afero.Fs, path string) (*Application, error) {
+func LoadFromFile(fs afero.Fs, path string, validate bool) (*Application, error) {
```
davemooreuws, line 12: Wonder if we should change this to skipValidate and validate by default?

jyecusch: There isn't really a default value for a func param. If we switch to an options struct that would make sense though

PR #852, opened 2025-07-04
Skips client generation if no client resources exist

jyecusch:
Prevents client generation when the application specification lacks compatible resources.

- Avoids unnecessary processing and potential errors by checking for client-adaptable resources (e.g., buckets) before initiating the client generation process.
- Improves efficiency by skipping client generation for applications that do not require it.

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -42,6 +42,11 @@ var generateCmd = &cobra.Command{
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


HomelessDinosaur:
Will more changes needed to be made the python SDK? Should we merge that in first?

HomelessDinosaur (approved):

PR #858, opened 2025-07-09
allow $PORT in dev commands

jyecusch:


raksiv (approved):
Looks good - Since this is a once off, its safe to assume other variables like $PORT_ABC won't exist.

PR #862, opened 2025-07-11
update output and write to terraform directory.

tjholm:


tjholm:
quick testing demo:
Uploading 2025-07-11 14-53-04.mp4…

cli/pkg/app/nitric.go
```diff
@@ -230,27 +232,54 @@ func (c *NitricApp) Build() error {
...
 	platform, err := terraform.PlatformFromId(fs, targetPlatform, platformRepository)
-	if err != nil {
+	if errors.Is(err, terraform.ErrUnauthenticated) {
+		fmt.Println("Please login first, using the `login` command")
+		return nil
+	} else if err != nil {
 		return err
 	}
 
+	fmt.Fprintf(os.Stdout, "\nBuilding application for %s\n\n", targetPlatform)
```
jyecusch, line 272: Note for future, I think we should probably attach an output writer to the `NitricApp` struct and refactor all the commands to write exclusively to that writer instead of fmt.Print/f/ln. Then in automated tests we can assert the outputs if we like.

tjholm: This is how the terraform engine is currently managing this using `FPrintf`

jyecusch (approved):

PR #864, opened 2025-07-14
allow listing all config keys and descriptions

jyecusch:
Allows user to list all available config keys with `nitric config -l` and the current values with `nitric config`.

Fixes NIT-74

### DIFF
--- a/cli/cmd/config.go
+++ b/cli/cmd/config.go
@@ -2,12 +2,30 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/nitrictech/nitric/cli/internal/config"
 	"github.com/nitrictech/nitric/cli/internal/style"
 	"github.com/samber/do/v2"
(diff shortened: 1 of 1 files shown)


jyecusch:
Some example usage of the config and config list options:

### When the current project has a local config file
```bash
nitric config
Config file: .nitric/config.yaml
=============================
url: http://localhost:3000
```

### Where there is a global config file
```bash
nitric config
Config file: /Users/jye/.nitric/config.yaml
=============================
url: http://localhost:3000
```

### When there is no config file
```bash
nitric config
Config file: none (using defaults)
=============================
url: https://app.nitric.io/ (default)
```

### Listing all keys
```bash
nitric config -l
Available configuration keys:
=============================
url: The base URL of the Nitric server (e.g., https://app.nitric.io)
```

davemooreuws (approved):

tjholm (approved):
Would probably make list a command instead of a flag, but happy with the implementation.

PR #866, opened 2025-07-15
feat: Add custom validation error messages to the build command and websocket sync

HomelessDinosaur:
Frontend PR https://github.com/nitrictech/platform/pull/75


tjholm (changes requested):

cli/pkg/schema/errors.go
```diff
@@ -0,0 +1,245 @@
...
+func FormatValidationErrors(errs []ValidationError) string {
+	if len(errs) == 0 {
+		return ""
+	}
+
+	arrow := "--->"
+	linePrefix := "  |"
+	var errsStr strings.Builder
+	for _, err := range errs {
+		errsStr.WriteString(fmt.Sprintf("%s %s\n", arrow, err.Message))
```
tjholm, line 235: strings builder also implements `io.Writer` so an alternative to calling Writing string and fmt.Sprinf is:

```suggestion
	fmt.Fprintf(&errsStr, "%s %s\n", arrow, err.Message)
```

cli/pkg/schema/errors.go
```diff
@@ -0,0 +1,245 @@
...
+	|
+	| YAML Context
+	|
+*/
+func FormatValidationErrors(errs []ValidationError) string {
+	if len(errs) == 0 {
+		return ""
+	}
+
+	arrow := "--->"
```
tjholm, line 231: const

cli/pkg/schema/errors.go
```diff
@@ -0,0 +1,245 @@
...
+	return "{{ invalid_property_name .field}} {{.property}}"
+}
+
+func (t *NitricErrorTemplate) DoesNotMatchPattern() string {
+	return "{{pretty_print_pattern .pattern}}"
+}
+
+func (t *NitricErrorTemplate) Required() string {
+	return "The {{.property}} property is required"
+}
```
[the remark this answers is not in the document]
tjholm: Ahh I see they're part of the `gojsonschema.locale` interface, might need to add some context/comments so that a bit clearer

cli/pkg/schema/yaml_context.go
```diff
@@ -0,0 +1,96 @@
...
+func (b *YamlContextBuilder) WriteError(s string) (int, error) {
+	errStr := lipgloss.NewStyle().Foreground(colors.Red).Render(fmt.Sprintf("\t# <-- %s", s))
+	return b.Builder.WriteString(errStr)
+}
+
+func (b *YamlContextBuilder) WriteYamlKey(s interface{}, indent int) (int, error) {
+	indentStr := strings.Repeat("  ", indent)
+	// If the key is an integer, it means its an array index, replace it with a dash.
+	keyStr := ""
+	if _, err := strconv.Atoi(fmt.Sprintf("%v", s)); err == nil {
```
tjholm, line 26: does `s` come through as its original type I'm assuming it does based on the arg.

A switch with type might be more suited here:

```go
switch v := s.(type) {
	case int:
		// v is an int
	case string:
		// v is a string
	default:
		// v is :shrug:
}
```

cli/internal/devserver/filesync.go
```diff
@@ -65,11 +67,39 @@ func (fs *NitricFileSync) setFileContents(contents []byte) error {
 	return err
 }
 
+func (fs *NitricFileSync) checkFileValidity(contents []byte) ([]schema.ValidationError, error) {
```
jyecusch, line 70: I think this name could be a little better, to make it clear what file it's referring to. E.g. `validateAppSchemaYaml` or something like that

jyecusch: It also look like this doesn't need to be attached to the struct (i.e. remove `(fs *NitricFileSync)`)

davemooreuws (approved):

cli/pkg/schema/validate.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+		resourceNames[name] = "website"
+	}
+
+	return violations
+}
+
+func (a *Application) checkNoReservedNames() []gojsonschema.ResultError {
+	violations := []gojsonschema.ResultError{}
+	reservedNames := []string{"backend"}
```
jyecusch, line 69: I think it's worth us documenting (with a comment for now), why any word is in this list. E.g. in this case due to terraform using backend as a reserved word.

It will give our future selves the info needed to reassess or not touch these values.

jyecusch (changes requested):

jyecusch (approved):

PR #867, opened 2025-07-15
Allow multiple dev commands to be running at the same time

HomelessDinosaur:
Fixes NIT-79


cli/internal/simulation/simulation.go
```diff
@@ -61,9 +66,16 @@ func (s *SimulationServer) startNitricApis() error {
...
+		openPort, err := netx.GetNextPort(netx.MinPort(NITRIC_SERVICE_MIN_PORT), netx.MaxPort(NITRIC_SERVICE_MAX_PORT))
+		if err != nil {
+			return fmt.Errorf("failed to find open port: %v", err)
+		}
+
+		port = fmt.Sprintf("%d", openPort)
 	}
 
 	addr := net.JoinHostPort(host, port)
+	os.Setenv("NITRIC_SERVICE_ADDRESS", addr)
```
jyecusch, line 78: You could probably attach this to the child processes directly (dev command). Otherwise I think we could have issues when dev cmd A starts some processes, then dev cmd B overwrites this value, then the dev service in A restarts

jyecusch (changes requested):

cli/internal/simulation/simulation.go
tjholm: Agree that it shoud be always attached to the sub process, but also os.Setenv is scoped to the environ of the current process only so it shouldn't be possible for seperate `dev` processes to overwrite each others `Environ` right?

tjholm:
Have pushed the requested changes, also added a public `ReservePort` method to netx to attempt to reserve the port if configured via an environment variable to remove it from the in process port pool.

tjholm (approved):

davemooreuws (approved):

PR #881, opened 2025-08-11
Add support for specifying a env_var_key as part of a database definition

tjholm:
also includes a crude breakup of the current apply function. Still needs more refactoring to increase test-ability.

### DIFF
--- a/cli/pkg/schema/database.go
+++ b/cli/pkg/schema/database.go
@@ -4,6 +4,8 @@ type DatabaseIntent struct {
 	Resource `json:",inline" yaml:",inline"`
 
 	Access map[string][]string `json:"access,omitempty" yaml:"access,omitempty"`
+	// The key that this database will export to services that depend on it.
+	EnvVarKey string `json:"env_var_key" yaml:"env_var_key"`
 }
 
 func (d *DatabaseIntent) GetAccess() (map[string][]string, bool) {
--- a/cli/pkg/schema/validate.go
+++ b/cli/pkg/schema/validate.go
@@ -8,12 +8,28 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 )
 
+func (a *Application) checkNoEnvVarCollisions() []gojsonschema.ResultError {
+	violations := []gojsonschema.ResultError{}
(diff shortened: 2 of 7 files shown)


cli/pkg/schema/database.go
```diff
@@ -4,6 +4,8 @@ type DatabaseIntent struct {
 	Resource `json:",inline" yaml:",inline"`
 
 	Access map[string][]string `json:"access,omitempty" yaml:"access,omitempty"`
+	// The key that this database will export to services that depend on it.
+	EnvVarKey string `json:"env_var_key" yaml:"env_var_key"`
```
tjholm, line 8: This is the main part of the change

HomelessDinosaur:
Looks good, saw no missed code in the refactor. Maybe we could split the `resources_handler` into separate functions per resource?

cli/pkg/schema/validate.go
```diff
@@ -8,12 +8,28 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 )
 
+func (a *Application) checkNoEnvVarCollisions() []gojsonschema.ResultError {
```
davemooreuws, line 11: Should this validation also apply during edit? It doesn't at the moment.

davemooreuws:
@tjholm needs a rebase

tjholm:
> Looks good, saw no missed code in the refactor. Maybe we could split the `resources_handler` into separate functions per resource?

you mean separate files? Yeah can do that, can also include this later (go code is pretty fluid when its the same package). I've rebased this branch a fair few times already

cli/pkg/schema/validate.go
tjholm: Looks like the validation on file edits ONLY ever runs if there is actually an error loading the yaml config. See filesync.go. This was to prevent broken config from reaching the UI. I'll raise a new issue to have additional validations included as part of that as currently none of them are because of this case.

I've tested it out locally with some changes but it has a much bigger impact on UX (completely blocking the user from using the UI at all! So I think this warrants a separate issue.

jyecusch (approved):

PR #885, opened 2025-08-13
Add team switching command

HomelessDinosaur:
Closes NIT-139

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/charmbracelet/lipgloss"
 	"github.com/nitrictech/nitric/cli/internal/config"
+	"github.com/nitrictech/nitric/cli/internal/style/colors"
 	"github.com/nitrictech/nitric/cli/internal/version"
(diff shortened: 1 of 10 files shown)


davemooreuws:
We should clean these errors up:

image

Steps:
1. Login and switch team
2. Revoke your session in workos
3. Try switching again

davemooreuws:
Is this supposed to change the workos org in the session/ui? If I go to the dashboard after switching teams, it is still the previous team.

jyecusch:
> Is this supposed to change the workos org in the session/ui? If I go to the dashboard after switching teams, it is still the previous team.

The easiest way to support that would probably be to move the `/dev` route behind a team slug `/{team_slug}/dev` so the UI auto-switches teams.

HomelessDinosaur:
> We should clean these errors up:
> 
> image
> Steps:
> 
> 1. Login and switch team
> 2. Revoke your session in workos
> 3. Try switching again

Screenshot 2025-08-15 at 4 54 35 PM

PR #888, opened 2025-08-15
Ensure redirects are passed through to the client and not handled.

tjholm:


runtime/service/proxy.go
```diff
@@ -14,12 +14,19 @@ type HttpServerProxy struct {
 	target string
 }
 
+var defaultHttpClient = &http.Client{
+	CheckRedirect: func(req *http.Request, via []*http.Request) error {
+		// Prevent redirects to ensure we handle them manually if needed
```
jyecusch, line 19:
```suggestion
		// Prevent redirects so the end user client can follow them itself
```

jyecusch (changes requested):

jyecusch (approved):

PR #889, opened 2025-09-09
fix: update github.com/pulumi/pulumi-awsx/sdk to v3

jyecusch:


jyecusch:
Thanks for the contribution @eshita090-a11y

PR #891, opened 2025-09-17
docs: excludes comparison docs from search indexing

davemooreuws:
Hides comparison documentation pages from search engine indexing and the sitemap.

This prevents these pages from appearing in search results.

The changes include:

- Adding a `noindex` field to the contentlayer configuration
- Setting `noindex: true` in the frontmatter of comparison MDX documents
- Updating the sitemap generation to exclude comparison pages
- Adding a disallow rule to the robots.txt for the `/misc/comparison/` path

Closes NIT-11

### DIFF
--- a/docs/contentlayer.config.ts
+++ b/docs/contentlayer.config.ts
@@ -39,6 +39,10 @@ const baseFields: FieldDefs = {
     type: 'string',
     description: 'The canonical url of the doc, if different from the url',
   },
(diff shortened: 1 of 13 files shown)


HomelessDinosaur:
Is this also removing it from the Algolia search? Not sure how that is indexed.

sean-nitric (approved):

